PR #2, opened 2021-03-11
feat: wrap initialize sdk function

rnike:
From issue facebook/react-native-fbsdk#840

I wrap the initialize function from native, the original pr is facebook/react-native-fbsdk#841

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,34 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also 
 
 ## Usage
 
+### SDK Initialization
+
+The `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900). 
+
+On iOS, developers are required to initialize the sdk after app launched, implement the code below in your `AppDelegate.m`:
+
+```objective-c
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
+{
+  [FBSDKApplicationDelegate initializeSDK:launchOptions]; // <- add this
+
+  // your other stuff
+}
+```
+
+Or call `Settings.initializeSDK` in anywhere else if you want the sdk to be initialized in react-native
+
+```js
+import { Platform } from 'react-native';
+import { Settings } from 'react-native-fbsdk-next';
+
+if(Platform.OS === 'ios'){
+  Settings.initializeSDK();
+}
+```
+
+> NOTE: `Settings.initializeSDK` is also functional for Android, if you want to comply with [GDPR](https://developers.facebook.com/docs/app-events/gdpr-compliance), `Settings.initializeSDK` can be called after an end user provides consent.
+
 ### [Login](https://developers.facebook.com/docs/facebook-login)
 
 #### Login Button + Access Token
--- a/README.md
+++ b/README.md
@@ -164,9 +164,27 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also 
 
 ### SDK Initialization
 
-The `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900). 
+To comply with Apple privacy requirements, for iOS the `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900).
 
-On iOS, developers are required to initialize the sdk after app launched, implement the code below in your `AppDelegate.m`:
+Using this module, there are two options to comply with this requirement, one is platform-neutral and can be used from your javascript code whenever it makes sense for your app, and one is native.
+
+#### Platform-neutral SDK initialization
+
+If you do not need to handle a [GDPR-type opt-in flow](https://developers.facebook.com/docs/app-events/gdpr-compliance), on iOS you should include the following javascript code as early in startup as possible. For Android auto-init is the default still, so this is not strictly necessary for Android but will work.
+
+If you need to handle a GDPR-type flow, make sure your SDK is configured natively to delay all logging activity according to [the GDPR instructions](https://developers.facebook.com/docs/app-events/gdpr-compliance), ask for user consent, and after obtaining consent include code similar to this:
+
+```js
+import { Settings } from 'react-native-fbsdk-next';
+
+// Ask for consent first if necessary
+// Possibly only do this for iOS if no need to handle a GDPR-type flow
+Settings.initializeSDK();
+```
+
+#### iOS Native Initialization
+
+If you would like to initialize the Facebook SDK even earlier in startup for iOS, you need to include this code in your AppDelegate.m file now that auto-initialization is removed.
 
 ```objective-c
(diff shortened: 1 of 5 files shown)


thebergamo:
I looks fine, but I believe it would be good to have some note in the `README.md` about it that you should do it before start using it.

mikehardy:
Yes! I believe this would stop the crash I saw and worked around by adding an AppDelegate chunk. +1 on a README note as otherwise I would have no idea, maybe even a major version bump

rnike:
README is added, any suggestions if it's not clear.

rnike:
@mikehardy I see, that's why you've added the code in `didFinishLaunchingWithOptions`.

README.md
```diff
@@ -162,6 +162,17 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also
...
+### SDK Initialization
+
+The `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900). On iOS, developers are required to initialize the sdk after app launched.
+
+```js
+import { Settings } from 'react-native-fbsdk';
+
+Settings.initializeSDK();
+```
+#### NOTE: This method is also functional on Android
```
thebergamo, line 174: I would just change the markdown here to use quote instead of h4.
Like this:

> NOTE: This method is also function on Android

README.md
```diff
@@ -162,6 +162,17 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also
 
 ## Usage
 
+### SDK Initialization
+
+The `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900). On iOS, developers are required to initialize the sdk after app launched.
+
+```js
+import { Settings } from 'react-native-fbsdk';
```
thebergamo, line 170: This is something that needs to be done earlier in the application or only when using FBSDK?
It's not that clear to me. 
For instance, I use FB only in my sign in page, can I do it only there or would it be better to do the initialization in my app.js?

thebergamo (changes requested):

thebergamo:
Besides comments, looks like there is some error in iOS build. 
@rnike Anyway, thanks for this PR, I totally forgot about it while just upgrading version.

@mikehardy I agree with you. this could be a major bump.

README.md
mikehardy: In my experience I think the app will login okay, but will crash if you run a graph request after? That was my experience when I did not put in my AppDelegate `didFinishLaunchingWithOptions` call to the FBSDK, and I **think** (unverifed / open question) `initializeSDK` would have the same effect.

I believe you would want to do it immediately unless you had a GDPR-style opt-in flow which you would have to satisfy prior to using a facebook login anyway I think?, then you would do it immediately after that

thebergamo: Yes, that was my point to maybe just to enforce the need to make it happening as earlier as possible in the implementation of their app. 

Unless of course you should obey some GDPR-style flow.

README.md
rnike: Sure, quoted text looks better for me.

README.md
rnike: @thebergamo I've got your point, let me complete the document.

rnike:
The `AutoInitEnabled` default is `true` if not set on Android. If following GDPR-style we will need to also wrap `FacebookSdk.setAutoInitEnabled` for Android ([reference](https://developers.facebook.com/docs/app-events/gdpr-compliance)).

In this pr I will be focused on sdk initialization.

thebergamo (approved):
I believe now everything looks good to me :)

rnike:
just accidentally unassigned myself, don't mine it

thebergamo:
Fixed :)

thebergamo:
Let's just wait for @mikehardy feedback on it and I'm good to merge and release it.

README.md
```diff
@@ -162,6 +162,34 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also
...
+import { Platform } from 'react-native';
+import { Settings } from 'react-native-fbsdk-next';
+
+if(Platform.OS === 'ios'){
+  Settings.initializeSDK();
+}
+```
+
+> NOTE: `Settings.initializeSDK` is also functional for Android, if you want to comply with [GDPR](https://developers.facebook.com/docs/app-events/gdpr-compliance), `Settings.initializeSDK` can be called after an end user provides consent.
+
```
mikehardy, line 192:
```suggestion

To comply with Apple privacy requirements, for iOS the `autoInitEnabled` option is removed from [facebook-ios-sdk#v9.0.0](https://github.com/facebook/facebook-ios-sdk/blob/master/CHANGELOG.md#900).

Using this module, here are two options to comply with this requirement, one is platform-neutral and can be used from your javascript code whenever it makes sense for your app, and one is native.

#### Platform-neutral SDK initialization

If you do not need to handle a [GDPR-type opt-in flow](https://developers.facebook.com/docs/app-events/gdpr-compliance), on iOS you should include the following javascript code as early in startup as possible. For Android auto-init is the default still, so this is not strictly necessary for Android but will work.

If you need to handle a GDPR-type flow, make sure your SDK is configured natively to delay all logging activity according to [the GDPR instructions](https://developers.facebook.com/docs/app-events/gdpr-compliance), ask for user consent, and after obtaining consent include code similar to this:

```js
import { Platform } from 'react-native';
import { Settings } from 'react-native-fbsdk-next';

// Possibly ask for consent first if necessary?
// Possibly only do this for iOS if Android is still set to auto initialize natively?
Settings.initializeSDK();
```

#### iOS Native Initialization

If you would like to initialize the Facebook SDK even earlier in startup for iOS, you need to include this code in your `AppDelegate.m` file now that auto-initialization is removed.

```objective-c
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  [FBSDKApplicationDelegate initializeSDK:launchOptions]; // <- add this

  // your other stuff
}
```

mikehardy (commented):
I have no problems with the code (it's easy code thankfully!) but I made a suggestion to restructure the document with the idea it will nudge users to be platform-neutral and use the Javascript code. And I expanded the information about GDPR vs non-GDPR flow as I think that is the real decision point, with most having to handle GDPR style flows (voluntarily or by law) so I paid special attention to directing people how to do that so it is the same on both platforms using the platform-neutral javascript call

Words are easy to smush around though, do anything you like (or nothing at all) with the suggestion :-)

README.md
mikehardy: Unfortunately the formatting of code blocks interferes with the suggestion blocks :-( - but all the markup is correct in there

rnike: Thanks a lot, I apply almost all of the suggestion except of the js example, here's the diff
```diff
- import { Platform } from 'react-native';
import { Settings } from 'react-native-fbsdk-next';

- // Possibly ask for consent first if necessary?
- // Possibly only do this for iOS if Android is still set to auto initialize natively?
+ // Ask for consent first if necessary
+ // Possibly only do this for iOS if no need to handle a GDPR-type flow
Settings.initializeSDK();
```

mikehardy: All looks good to me! Cheers @rnike

thebergamo:
As you see, release done :)

thebergamo:
Thank you very much @rnike for your PR and the commitment into it that was wonderful and @mikehardy for reviewing and providing valuable feedback on it as well.

Issue #6, opened 2021-03-18
I think typescript is useful for this project


jihokim2:
Hello, Do you have plan to make index.d.ts?

thebergamo:
Hi @jihokim2 yes, it's part of the migration improvement. 
If you have some experience with Typescript, feel free to open a PR with this .d.ts while migration is still in progress :)

thebergamo:
work is getting done here #10

thebergamo:
TS definititons are released with version 4.1.0
https://github.com/thebergamo/react-native-fbsdk-next/releases/tag/v4.1.0

PR #10, opened 2021-04-06
chore: Add typescript type definitions

g-dimitry:
Hello,

Due to the power of Typescript, it is widely used by the community (including me). So to support all Typescript users, I generated typescript type definitions for all current modules.

Steps taken to generate:
1) Use [babel-plugin-flow-to-typescript](https://www.npmjs.com/package/babel-plugin-flow-to-typescript) plugin to transform flow to typescript
2) Transform all require statements into import statements
3) Fix all module.exports into typescript exports
4) Generate type definitions using tsc
4) Remove the declare keyword from all files but index.d.ts
5) Declare module with the same name as the package inside index.d.ts
6) Edit all import statements from "import Foo from './bar'" to "import('./bar').default"

Steps taken to add into package:
1) Create a directory called "types" at the root of the repo
2) Add all type definitions files inside it
3) Add "types" directory under "files" inside package.json
4) Add property "types" inside package.json with value "types/index.d.ts"

Steps taken to test:
1) Install package using git inside another project
2) Check that types are accessible

Credit for the generation go to the babel-plugin-flow-to-typescript babel plugin and this medium [article](https://www.npmjs.com/package/babel-plugin-flow-to-typescript)

Thank you


thebergamo:
@g-dimitry thank you very much for your PR adding those TS type definitions to this project. 

Even as I understand that the code is auto generated, but could we get rid of this strange exports and also the eslint disable comments?

g-dimitry:
Yes sure the eslint disable lines are my bad I'll remove them right away. And to be sure, what you mean by the strange exports are the
`const defaultExport: {}
export = defaultExport;
`

thebergamo:
Exactly these ones hehe

thebergamo (commented):
As per current changes I would say that's ok. There are few more tweeks that I would do for example, this styles that basically are duplicated in few places and they're basically the same object being exported many times.
But not a blocker to merge it.

In my evening I will merge and try to push it to register the types

types/FBShareButton.d.ts
```diff
@@ -1,6 +1,3 @@
-/* eslint-disable @typescript-eslint/no-use-before-define */
```
thebergamo: Didn't your linter complain about `styles` being used before it's declared?

micaww:
this looks like a good solution for now. @thebergamo is the end goal to migrate the code away from flow to typescript? i would like to start working on that if so.

g-dimitry:
Guys I'm sorry for the late response this code isn't ready I found a little problem with it and I'm currently trying to fix it. Please refrain from merging until it's fixed. And yes if we can migrate from flow to typescript that would be awesome. It can be done automatically and I can do it if you want in another pull request.

types/FBShareButton.d.ts
g-dimitry: Yes but this actually a typescript issue and not a type declarations error

g-dimitry:
@thebergamo Type defs are done. I checked on all exported modules. Please double check and merge. And of course if you have any other comments or concerns please let me know.

g-dimitry:
> this looks like a good solution!
> 
> @thebergamo is the end goal to migrate the code away from flow to typescript? i would like to start working on that if so.

I definitely support the migration from flow to typescript it's more widely used. The migration is automatic although some manual changes will be made as pointed out by the steps to produce the type defs above. I'll be happy to do it, if this was agreed on of course, as I already had to turn flow to TS to generate the type defs.

micaww:
> 
> 
> I'll be happy to do it, if this was agreed on of course, as I already had to turn flow to TS to generate the type defs.

can't speak for the maintainer but seems like making the source typescript is part of the plan according to this comment on the issue: https://github.com/thebergamo/react-native-fbsdk-next/issues/6#issuecomment-801775192

definitely would be awesome if the source was typescript and the definitions didn't have to be updated manually after the first time. there isn't much JS in this library so should be pretty easy to maintain as well

thebergamo:
Yes, the end goal is the migration to TS, and thank you again for the typedefs I'm in a bit hurry with other personal things but I will definitely test it today and will publish a new version o/

thebergamo:
@g-dimitry everything looks to me! I've downloaded your branch and it points to the types properly in my vs-code! I will include it in the next release. I believe it would be beneficial to have the PR #15 before releasing a new version anyway :)

Thanks again for your time and effort in adding those types!

PR #17, opened 2021-04-15
fix: Support AccessToken nullable fields

superandrew213:
`AccessToken` requires only `accessToken`, `applicationId` and `userId`. See: https://github.com/facebook/facebook-android-sdk/blob/master/facebook-core/src/main/java/com/facebook/AccessToken.java#L111-L121. Other fields are nullable.

Fixes https://github.com/facebookarchive/react-native-fbsdk/issues/456.

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
@@ -60,22 +60,17 @@ import androidx.annotation.Nullable;
 public final class Utility {
 
     public static AccessToken buildAccessToken(ReadableMap accessTokenMap) {
-        AccessTokenSource accessTokenSource = AccessTokenSource
-                .valueOf(accessTokenMap.getString("accessTokenSource"));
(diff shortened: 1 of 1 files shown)


thebergamo:
Code looks good to me! Thanks for contributing @superandrew213

se09deluca:
Hi, 
my app is running on the actually latest version of this package (4.2.0).

Some of my users still experience this error. Unfortunately i cannot reproduce this error on my side. 
But i can attach the stacktrace from Sentry

null pointer exception facebook sdk


This bug is affecting near to 1k users.  😩

se09deluca:
Hi @mikehardy , thanks for the reply. I haven't replied with a repro because i cannot reproduce the error (it works on my machine 😆). 
After some digging in the native code i notice that accessToken is an optional field, the native SDK (if i have understood correctly) would self resolve it. 
After a week without any errors in our A/B test i would say that the bug in my code is resolved. 
I've read erroneously the documentation from facebook.

Issue #7, opened 2021-03-18
Expose the new LoginManager -> limited login tracking functionality.


Techdojo:
Looking through the code it would appear that the new limited login functionality available via the LoginManager and or LoginButton has not yet been exposed to React Native.

Is anyone already working on this or have a timeline for this. Alternatively I'd be happy to help as it looks like it should be a fairly straightforward process.
[https://developers.facebook.com/docs/facebook-login/limited-login/ios/](url)

thebergamo:
Hi @Techdojo thanks for your suggestion! At this moment I don't have any specific plan towards it.

As you suggested, feel free to contribute with this new lib, I will be more than happy to review and merge it for future release :)

Techdojo:
Thanks, I'll have a look at what's involved. 🤞 it shouldn't be a massive change.

thebergamo:
No biggie :)
If you see any space that we can work together on that, let me know and we can split up the work

leolusoli:
Hi guys.
I've made a fork adding a separate (and very simple) limitedLogin method in order to add the desired behaviour. Before opening the PR, I'd like to know if someone else are making the same part in order to organise the work. Thanks in advice :)

mikehardy:
@leolusoli fantastic! No one else has done anything in the area that I know of, post that PR :smile: !

aganov:
With a bit of modifications I was able to obtain `AuthenticationToken` but it is kind of useless. Only supported response type is jwt encoded `id_token` https://www.facebook.com/.well-known/openid-configuration/ which you are receiving in the app, and no way to safely send this to the server and replace the classic auth token. If this was intended to be a useful replacement, they should support `code` response type (like Apple [did](https://appleid.apple.com/.well-known/openid-configuration/) with their Apple ID). And with this `code` you can safely receive the user's `id`, `name` and `email` server side... Does anyone come to a different conclusion?

cocoa-rum:
@mikehardy Hi. I am also forked and implemented Limited Log in, but only for LoginManager. 
I can try to PR. Also you can check in my repo. Branch name [limited-login](https://github.com/cocoa-rum/react-native-fbsdk-next/tree/limited-login)

tl dr
created separated method **limitedLogInWithPermissions**
and for getting Authentication Token - class AuthenticationToken

leolusoli:
Hi @cocoa-rum. I'm watching your modifications. Can you give an example of using this method in JS? I also have some questions about it:

[...]

PR #18, opened 2021-04-15
feat: Limited login and Profile for IOS with type def

cocoa-rum:
PR for https://github.com/thebergamo/react-native-fbsdk-next/issues/7

Thanks  [@leolusoli](https://github.com/leolusoli)  for **_Profile_** part

PR contains Authentication Token along side with limited log in:
If we use limited login, we lose access to access token, so we need an alternative and Authentication Token is. So from my view point Authentication Token complements limited log in

Also Profile part
Related to the [commnent](https://github.com/thebergamo/react-native-fbsdk-next/issues/7#issuecomment-819483813)


src/FBAuthenticationToken.js
```diff
@@ -0,0 +1,64 @@
...
+     Object.freeze(this);
+   }
+ 
+   /**
+    * Getter for the authentication token
+    */
+   static getAuthenticationToken(): Promise<?FBAuthenticationToken> {
+     if (Platform.OS === 'android') {
+       return Promise.resolve(null);
+      }
```
bernard-kms, line 50: I understand AuthenticationToken only valid in iOS, but I think we can give more information to the developers about this platform-specific function via name or warnings. Something like `-IOS` suffix perhaps? e.g. `getAuthenticationTokenIOS`

src/FBProfile.js
```diff
@@ -0,0 +1,103 @@
...
+    this.name = profileMap.name;
+    Object.freeze(this);
+  }
+
+  /**
+  * Getter the current logged profile
+  */
+  static getCurrentProfile(): Promise<?FBProfile> {
+    if (Platform.OS === 'android') {
+      // TODO: Not implemented in Android yet
```
bernard-kms, line 86: I think this could be somewhat unexpected as Profile functionality is also implemented in Android SDK: https://developers.facebook.com/docs/reference/androidsdk/current/facebook/com/facebook/profile.html

I think it would be nice to implement Android side as well.

bernard-kms (commented):
I just leave some thoughts about platform-specific behavior. I think it would be more safe with careful implementation because not everyone well understand about each platform's limitations, especially in cross-platform libraries like react native.

cocoa-rum:
@bernard-kms I agree. I will improve what you said
Speaking about **_"platform-specific behavior"_**. Log in button contains `nonce` and `loginTracking`, which are specifice for IOS too, should I rename them `nonceIOS` and `loginTrackingIOS` as well ?

bernard-kms:
> Speaking about **_"platform-specific behavior"_**. Log in button contains `nonce` and `loginTracking`, which are specifice for IOS too, should I rename them `nonceIOS` and `loginTrackingIOS` as well ?

I think that would be safer(maybe the safest) way to implement it. When limited login available in android (FB would implement it, wouldn't they?) we can add a new cross-platform one while gradually deprecating the platform-specific one and not making any breaking changes.

src/FBProfile.js
leolusoli: I'm doing it right now.

cocoa-rum:
@bernard-kms @leolusoli 
I think we are ready for review

_Also, what the best practice with package. I think, I am doing it a little bit complex way. It's my first time working with external package in such way
I added forked repo ( yarn add *local repo path*) Into my project. Do edits and paste changes into the package - commit_

ios/RCTFBSDK/login/RCTFBSDKLoginButtonManager.m
```diff
@@ -45,6 +45,14 @@ - (UIView *)view
 
 RCT_EXPORT_VIEW_PROPERTY(defaultAudience, FBSDKDefaultAudience)
 
+ RCT_CUSTOM_VIEW_PROPERTY(nonceIOS, NSString, FBSDKLoginButton){
```
glenna, line 48: style nit for consistency, can you put the opening bracket on a new line 🙏🏻

ios/RCTFBSDK/login/RCTFBSDKLoginButtonManager.m
```diff
@@ -45,6 +45,14 @@ - (UIView *)view
...
+   [view setNonce:json ? json : nil];
+ }
+
+ RCT_CUSTOM_VIEW_PROPERTY(loginTrackingIOS, NSString, FBSDKLoginButton){
```
glenna, line 52: same here :)

ios/RCTFBSDK/login/RCTFBSDKLoginButtonManager.m
```diff
@@ -45,6 +45,14 @@ - (UIView *)view
...
+   [view setLoginTracking:(json && [json  isEqual: @"limited"]) ? FBSDKLoginTrackingLimited : FBSDKLoginTrackingEnabled];
```
glenna, line 53: Here, since we know `json` is a string, it is faster to use `-[NSString isEqualToString:]` method, so `[json isEqualToString:@"limited"]`

ios/RCTFBSDK/login/RCTFBSDKLoginManager.m
```diff
@@ -63,18 +63,39 @@ + (BOOL)requiresMainQueueSetup
...
-      reject(@"FacebookSDK", @"Login Failed", error);
+    FBSDKLoginManagerLoginResultBlock requestHandler = ^(FBSDKLoginManagerLoginResult *result, NSError *error) {
+        if (error) {
+            reject(@"FacebookSDK", @"Login Failed", error);
+        } else {
+            resolve(RCTBuildResultDictionary(result));
+        }
+    };
+    FBSDKLoginConfiguration *configuration;
+    FBSDKLoginTracking tracking = [loginTracking  isEqual: @"limited"] ? FBSDKLoginTrackingLimited : FBSDKLoginTrackingEnabled;
```
glenna, line 79: Same as above: `-[NSString isEqualToString:]`

src/FBProfile.js
```diff
@@ -0,0 +1,96 @@
...
+  middleName: string,
+  imageURL: string,
+  linkURL: string,
+  userID: string,
+  email: string,
+  name: string,
+};
+
+/**
+ * Rappresent the current profile logged
```
glenna, line 23: I think this is a typo?

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
```diff
@@ -100,6 +101,69 @@ public static WritableMap accessTokenToReactMap(AccessToken accessToken) {
...
+        String userId = profile.getId();
+        if (userId == null) {
+            map.putNull("userID");
+        } else {
+            map.putString("userID", userId);
+        }
+
+        // NOTE: Email doesn't get retrieved directly by the profile instance after a login. You have to fetch the graph API in
+        // order to get this value.
+        map.putNull("email");
```
thebergamo, line 155: why do we need to place email here if it doesn't come from the endpoint itself?

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
```diff
@@ -100,6 +101,69 @@ public static WritableMap accessTokenToReactMap(AccessToken accessToken) {
...
+
+        String name = profile.getName();
```
thebergamo, line 157: I would just move this up as to have the data grouped in a logical sort, as you've the `firstName`, `middleName` and `lastName` on top

src/FBLoginManager.js
```diff
@@ -61,11 +61,26 @@ export type LoginResult = {
...
   /**
-   * Logs the user in with the requested permissions.
+   * Log in with the requested permissions.
+   * @param loginTracking IOS only: loginTracking: 'enabled' | 'limited', default 'enabled'.
+   * @param nonce IOS only: Nonce that the configuration was created with. A unique nonce will be used if none is provided to the factory method.
    */
-  logInWithPermissions(permissions: Array<string>): Promise<LoginResult> {
+  logInWithPermissions(
+    permissions: Array<string>,
+    loginTracking?: LoginTracking,
```
glenna, line 74: I see it is not standard in this lib, but other libs I work with also append the platform to these arguments to make it more clear that they are not used for all platforms. So similar as was done for the properties for FBLoginButton. But, as I'm not the maintainer here 🤷🏻‍♀️

ios/RCTFBSDK/core/RCTFBSDKProfile.m
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    return @{
+        @"firstName": FBSDKProfile.currentProfile.firstName ? FBSDKProfile.currentProfile.firstName : [NSNull null],
+        @"lastName": FBSDKProfile.currentProfile.lastName ? FBSDKProfile.currentProfile.lastName : [NSNull null],
+        @"middleName": FBSDKProfile.currentProfile.middleName ? FBSDKProfile.currentProfile.middleName : [NSNull null],
+        @"imageURL": FBSDKProfile.currentProfile.imageURL ? FBSDKProfile.currentProfile.imageURL.relativeString : [NSNull null],
+        @"linkURL": FBSDKProfile.currentProfile.linkURL ? FBSDKProfile.currentProfile.linkURL.relativeString : [NSNull null],
+        @"userID": FBSDKProfile.currentProfile.userID ? FBSDKProfile.currentProfile.userID : [NSNull null],
+        @"email": FBSDKProfile.currentProfile.email ? FBSDKProfile.currentProfile.email : [NSNull null],
```
thebergamo, line 36: On Android looks like we don't have this information coming from currentProfile, does iOS behave differently in such case? If yes, it would be nice to state it in the docs that only iOS should have value for this property.

ios/RCTFBSDK/core/RCTFBSDKProfile.m
```diff
@@ -0,0 +1,41 @@
...
+        @"name": FBSDKProfile.currentProfile.name ? FBSDKProfile.currentProfile.name : [NSNull null],
```
thebergamo, line 37: Again, maybe moving this prop to the top of this return

types/FBProfile.d.ts
```diff
@@ -0,0 +1,67 @@
...
+export class FBProfile {
+  
+  /**
+   * The user id
+   */
+  userID?: string;
+
+  /**
+   * The user's email.
+   * IMPORTANT: This field will only be populated if your user has granted your application the 'email' permission.
```
glenna, line 24: According to the comments in the (Android) impl file, you actually cannot retrieve it via the profile, or?

thebergamo (commented):
Besides my comments, there is also some warning about unused Platform variable!
Anyway, thanks very much for your contribution. Would like to see some comments from @bernard-kms if possible before integrate this with master ;)

glenna (commented):
Hey @cocoa-rum, thanks for writing this feature! I'm definitely looking forward to this being released 😄 

I'm new to this repo/conversation, but thought I'd also provide a review since my team will be using this feature soon.

For me, it looks good overall, just some minor style nits, to keep consistency, and another naming convention that probably @thebergamo would like to weigh in on.

Also, just a small comment that on reviewing I did not understand why Profile and Authentication Token was added as part of the Limited Login functionality, until I went through and read the comments in the issue, so it would be good in the future to make that more apparently in the PR description as well.

Thank you adding this functionality! Looking forward to the release ;)

src/FBLoginManager.js
thebergamo: I totally agree with your point here :)

ios/RCTFBSDK/core/RCTFBSDKProfile.m
glenna: Good catch, I didn't catch this one

cocoa-rum:
@thebergamo I got it. I will fix things that @glenna mentioned

> Also, just a small comment that on reviewing I did not understand why Profile and Authentication Token was added as part of the Limited Login functionality

As for Authentication Token - if we use limited login, we lose access to access token, so we need alternative and Authentication Token is. So from my view point Authentication Token complements limited log in

As for Profile, this [commnent](https://github.com/thebergamo/react-native-fbsdk-next/issues/7#issuecomment-819483813) 
And it appeared in this PR.
May be  we should ask @leolusoli  to create separate PR with Profile part? and from this PR we will remove it

src/FBLoginManager.js
cocoa-rum: What will be best practice for this?
Create **_logInWithPermissionsIOS_** and **_logInWithPermissionsAndroid_**

or something like this 
`logInWithPermissions ( { permissions,  loginTrackingIOS, nonceIOS } ) `

or did you mean
```
logInWithPermissions(
    permissions: Array<string>,
    loginTrackingIOS?: LoginTracking,
    nonceIOS?: string,
  )
```

glenna: @cocoa-rum

I mean the latter. So:

```
logInWithPermissions(
    permissions: Array<string>,
    loginTrackingIOS?: LoginTracking,
    nonceIOS?: string,
  )
```

glenna:
@cocoa-rum thank you! yeah, I got the context when I read the issue, but as a reviewer I don't expect to need to read all comments on an issue thread to get the context ;). Just a future reference to add more information to the PR description.

Regarding moving Profile impl to a separate PR, I would leave that up to @thebergamo. (If it were me, I would leave it in, as now it's reviewed and ready to roll -- but update the description so that in the release notes for the version when the PR is linked there's easy to access information)

thebergamo:
Regarding Profile impl, I believe it's pretty ok having it here, so no biggie from my side. You could just give more context as mentioned by @glenna in the PR description, so it would be much easier to track in the changelog later on :)

leolusoli:
@glenna @thebergamo I'll explain you tomorrow with some examples of what happening for Profile **imageURL** field, because there's some particular situation regarding limited/not limited login (i.e. different value returned). I'll also fix the typos in comments and style some code according to the guidelines described in the comments. I think tomorrow will be all pushed and explained :)

Thanks for your time again.

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
leolusoli: Because the Profile class inside the Java SDK doesn't expose the email field. I place email=null for maintain the same behaviour with iOS. Do you think that is more correct to move this concept in JS?

ios/RCTFBSDK/core/RCTFBSDKProfile.m
leolusoli: I recognise now that even the iOS sdk retrieve the url with width and height (i.e. https://platform-lookaside.fbsbx.com/platform/profilepic/?asid=10224600543422918&height=100&width=100&....).
The iOS SDK set two default width and height values for the profile picture (100). In Android, we have only the method `getProfilePictureUri` with two params **mandatory**. So, I think that the correct way to handling this part is to pass 100 (as a default size setted by iOS) to the `getProfilePictureUri` in order to retrieve the same image for the two platforms. In the future (I think is out of scope of this PR), we can add a method that get the profile picture with the width and height parameters (even in iOS we have the possibility to make this). Could be the right way?

types/FBProfile.d.ts
leolusoli: No, the email property is not inside the Profile implementation of the Android ([https://developers.facebook.com/docs/reference/androidsdk/current/facebook/com/facebook/profile.html/?locale=it_IT](url)). The only way to retrieve the email of the user is over GraphRequest. Maybe when Facebook will add the limited login in Android, they also add the email field in the Profile? Honestly, I don't know why this field is missing in Android.

bernard-kms: I think `-IOS` suffix could be also applied to this one to inform about limitation in Android.

cocoa-rum:
@thebergamo looks like we fixed all remarks for now
You are welcome to review or make some comments for improving PR

types/FBProfile.d.ts
leolusoli: Sorry @bernard-kms, can you tell me where to place this suffix please?

bernard-kms: I mean the `email` field, since this field is only valid for iOS as you mentioned.

leolusoli: Mmmm... Honestly, i don't think this is a correct tip because, if you are don't asking the correct permissions (i.e. the "email" permission) for the method **loginInWithPermissions** you can't retrieve the value in the field even in iOS. I think that the values of the Profile are dependent to the type of the auth request (and his options) and not to the platform.

ios/RCTFBSDK/core/RCTFBSDKProfile.m
thebergamo: Instead of having default values I would just stick with them as mandatory and later on when we implement the method to retrieve state it clearly.
And yes, I believe it can be part of a new PR, maybe you could please just create an issue and link it here to not lose track of this `getProfilePictureUri` method as someone else could pick it :)

types/FBProfile.d.ts
thebergamo: In that case we should have the value also on Android right? In such case I would stick with the suffix till we can maintain the same behavior in both platforms.

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
```diff
@@ -100,6 +101,61 @@ public static WritableMap accessTokenToReactMap(AccessToken accessToken) {
         return map;
     }
 
+    public static WritableMap profileToReactMap(Profile profile) {
```
thebergamo, line 104: If you could, just in case would be nice to have some simple utility to putNull or value into the WritableMap to avoid repeating yourself that much in this case as you're having too many `if/else` similarly here.

thebergamo (approved):
From my perspective it's fine as it is, just minor changes that we can do after releasing this version. Although the email that comes only from iOS I would a more clear note in the `readme.md` about this feature and this property comes only from iOS for now as it's not supported by Android.
After having confirmation from others reviewers (cc @bernard-kms && @glenna) and feature documented I'm good to merge and release it :)

glenna (commented):
Looking good!

I agree with @thebergamo about a section in the readme as well :)

types/FBLoginManager.d.ts
```diff
@@ -24,11 +24,20 @@ export type LoginResult = {
...
-     */
-  logInWithPermissions(permissions: Array<string>): Promise<LoginResult>;
+   * Log in with the requested permissions.
+   * @param loginTracking IOS only: loginTracking: 'enabled' | 'limited', default 'enabled'.
+   * @param nonce IOS only: Nonce that the configuration was created with. A unique nonce will be used if none is provided to the factory method. 
+   */
+   logInWithPermissions(
+      permissions: Array<string>,
+      loginTracking?: LoginTracking,
+      nonce?: string
```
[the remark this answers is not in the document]
cocoa-rum: Done. Even don't know, how I missed it

glenna:
@thebergamo did you also want the readme section before merging? If so, I think that's the last thing we are waiting on.

thebergamo:
Yes, because without that could lead others to some mistakes

glenna:
@thebergamo agreed! Just wanted to make it clear :)

cocoa-rum:
> Yes, because without that could lead others to some mistakes

Should I and @leolusoli  add changes to readme? Or what is next step?

thebergamo:
Yes, you should.

bernard-kms:
LGTM, and yes, a proper documentation is important to complete this work to be released.

cocoa-rum:
Readme is ready. Is it correct?

glenna (approved):
looks good to me! Thank you! 🙏🏻

thebergamo:
Version released :) [v4.2.0](https://github.com/thebergamo/react-native-fbsdk-next/releases/tag/v4.2.0)

Thank y'all for the effort for make it happen :)

PR #28, opened 2021-04-30
docs(README): update link to upstream iOS Getting Started Guide

birdwell:
The old link is broken. This updates to the correct link. We might change the step numbers because step 3 isn't reflected in the guide for some reason.


mikehardy (approved):
verified - thanks!

PR #31, opened 2021-05-06
fix: compileDebugJavaWithJavac

UriMiranda:
Propose to fix a build error in android. Missing constructor on FacebookProfile.

If this PR fixes an issue, type "Fixes #issueNumber" to automatically close the issue when the PR is merged.
thebergamo#30

Test Plan:
In a project with react-native-fbsdk-next as dependency enter in android folder and run:
cd android && ./gradlew app:installDebug -PreactNativeDevServerPort=8081

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBSDKPackage.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBSDKPackage.java
@@ -43,7 +43,7 @@ public class FBSDKPackage implements ReactPackage {
                 new FBGraphRequestModule(reactContext),
                 new FBLoginManagerModule(reactContext, mActivityEventListener),
                 new FBMessageDialogModule(reactContext, mActivityEventListener),
-                new FBProfileModule(),
+                new FBProfileModule(reactContext),
                 new FBSettingsModule(),
                 new FBShareDialogModule(reactContext, mActivityEventListener)
         );
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
@@ -16,6 +16,10 @@ import androidx.annotation.NonNull;
 public class FBProfileModule extends ReactContextBaseJavaModule {
     public static final String NAME = "FBProfile";
 
+    public FBProfileModule(@NonNull ReactApplicationContext reactContext) {
+      super(reactContext);
+    }
+
     @NonNull
     @Override
     public String getName() {
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
@@ -5,6 +5,7 @@ import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.module.annotations.ReactModule;
 
 import androidx.annotation.NonNull;
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
@@ -16,7 +16,7 @@ import androidx.annotation.NonNull;
 public class FBProfileModule extends ReactContextBaseJavaModule {
     public static final String NAME = "FBProfile";
 
-    public FBProfileModule(@NonNull ReactApplicationContext reactContext) {
+    public FBProfileModule(ReactApplicationContext reactContext) {
       super(reactContext);
     }
 


mikehardy:
Hi there! 

This should not be merged without an adequate explanation why the build succeeds in CI: https://github.com/thebergamo/react-native-fbsdk-next/runs/2491258907?check_suite_focus=true

If CI is successfully compiling - the assumption must be that

1) there is a project configuration error somehow or
2) that CI is somehow incorrect and not detecting something

If CI is somehow missing something, that's important to know and the PR should include a change to this workflow file to show the failure, and then the PR should include the fix and this should merge - https://github.com/thebergamo/react-native-fbsdk-next/blob/master/.github/workflows/ci.yml

mikehardy (changes requested):
I love fixing compile errors, but as mentioned in my comment on the PR itself this should not be merged until there is a diagnosis on why CI was building successfully vs your experience of having a build failure

mikehardy:
Would love to fix it: we must know why the example compiles in CI and your build fails first though

UriMiranda:
I think that is Grande version, is the only thing that changes from my machine to CI

UriMiranda:
But don't make sense, is just a simple constructor

mikehardy:
@UriMiranda okay: can you *try* it in your project with the identical version to verify? Then we will know and we can make real decisions on how to handle

UriMiranda:
Sure, I will try

UriMiranda:
Is not Grandle version, I tried with the same version. Do u know what is the java version? Here is 1.8

thebergamo:
From what I see here it uses Java 11:
[ubuntu-latest](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md)

mikehardy:
You don't have to guess, you can look at the action output.

For instance, you are attempting to fix a compile error but your PR does not actually compile: https://github.com/thebergamo/react-native-fbsdk-next/runs/2529266420?check_suite_focus=true - check the output and you see missing symbol

And the environment in general:
https://github.com/thebergamo/react-native-fbsdk-next/runs/2529266420?check_suite_focus=true
https://github.com/actions/virtual-environments/blob/ubuntu20/20210425.1/images/linux/Ubuntu2004-README.md
Yes Java 11 is the default on the runner and CI is using that. In fact it is the default for Android Studio 4.2 (came out this week) now.

I just made sure my project is using 4.2.0 here and JDK8, it works fine too.

With no reproduction, there's no issue I think.

mikehardy (changes requested):
Fails to compile in CI, project works without this in my project - need demonstration of failure in CI prior to merge if merged

mikehardy:
That explains it! An explanation is important.

So the default constructor was added for react-native 0.62 - https://github.com/facebook/react-native/commit/e69be0ae555dad2bdec0ea7e4fc62a4ee349cc2d

That means a couple things at once:

- the CHANGELOG can get an update for the already-released version indicating it has a requirement for minimum react-native version of 0.62 as an accidental breaking change
- this repo can make a decision on what minimum version of react-native should be supported (0.60? Does it even correctly operate on ios14.5? Since you need Xcode 12.5 so sign apps for ios14.5 and Xcode 12.5 won't compile react-native 0.62 even without patching, I think not)
- based on the decision of minimum react-native version support an issue can be logged here to have a separate CI task that does a test build against that minimum

With respect + understanding that it is not easy to update, I don't see react-native below 0.63 as viable even as that is the oldest one that was patched to work with Xcode 12.5 and that's a requirement for iOS development.

So I personally don't see react-native 0.60 / 0.61 compatibility as a reasonable target and I'd simply note the same in package.json / CHANGELOG / re-issue as a breaking change and counsel use of https://react-native-community.github.io/upgrade-helper/ so people get up to date. That is opinion, of course, but the Xcode 12.5 build thing is fact https://github.com/facebook/react-native/issues/31179

devdigitalrepublic:
I think the error has been caused by a annotation @NotNull in constructor, so a I removed. Other classes have the same constructor without it.

UriMiranda:
Hi, I guess the error is because of an annotation in construct (@NotNull), I saw others classes in code that has the same constructor without it.

UriMiranda:
I removed it, 50% of chance that will work. lol

android/src/main/java/com/facebook/reactnative/androidsdk/FBSDKPackage.java
```diff
@@ -43,7 +43,7 @@
                 new FBGraphRequestModule(reactContext),
                 new FBLoginManagerModule(reactContext, mActivityEventListener),
                 new FBMessageDialogModule(reactContext, mActivityEventListener),
-                new FBProfileModule(),
+                new FBProfileModule(reactContext),
                 new FBSettingsModule(),
```
mikehardy, line 47: isn't FBSettingsModule going to have the same issue :thinking:

UriMiranda: Good question 🤔

UriMiranda:
It works now!! Revolve ReactApplicationContext import.

mikehardy (approved):
I'm running CI on it now, and now that we know it's a react-version problem (so the problem is understood!) I'm +1 to get people moving.
I still argue (I think reasonably, with technical merit) that react-native 0.63 should be the minimum supported version going forward, but that's something that needs to be stated explicitly, and it has not been (yet)

thebergamo:
I'm fine supporting only version 0.63 of react-native as per it won't compile in newer Xcode.

When I get in my computer I will add a notice in the readme stating this in this PR.

thebergamo (approved):

PR #42, opened 2021-05-25
fix: update React-Core pod

philly25:
This is apparently needed for Xcode 12 compatibility

Reference: facebook/react-native#29633

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
   ],
   "dependencies": {},
   "peerDependencies": {
-    "react-native": ">=0.60.2"
+    "react-native": ">=0.63.3"
   },
   "devDependencies": {
     "@babel/core": "^7.12.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -71,7 +71,7 @@
   ],
   "dependencies": {},
   "peerDependencies": {
-    "react-native": ">=0.59.0"
+    "react-native": ">=0.60.2"
```
mikehardy, line 74: There is an unrelated issue that makes this better as an even higher peer dependency now:
```suggestion
    "react-native": ">=0.63.3"
```

(back story: Xcode 12.5 broken react-native and 0.63.3 is the oldest version patched so it's the oldest version still compatible with Xcode really)

mikehardy (approved):
Yep - this is the same one patched in to most of the modules I work on - just one unrelated nit to bump the react-native peerdep even higher, if @thebergamo agrees?

thebergamo (approved):

PR #47, opened 2021-05-28
build(android): remove build tools version

dulmandakh:
Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,7 +21,6 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 28)
-    buildToolsVersion safeExtGet('buildToolsVersion', '28.0.3')
     defaultConfig {
         minSdkVersion safeExtGet('minSdkVersion', 16)
         targetSdkVersion safeExtGet('targetSdkVersion', 28)


mikehardy (approved):
Yep

mikehardy:
Note that even if the PR title is fixed to be conventional, single-commit merges typically take the *commit* which is still possibly non-conventional. Since we use semantic-release bot you must be careful with the merge message! It will probably need a little manual fixup

dulmandakh:
Thanks @mikehardy

thebergamo (approved):
Lgtm

thebergamo:
That's was important tip, usually I the commit was taken from merge itself. It's a good point to make sure those commits also follow semantic release, will check if I can configure husky to avoid such issues

mikehardy:
alternatively there is a mode in the checker so it will flag the thing for more attention: 

```

          # When using "Squash and merge" on a PR with only one commit, GitHub
          # will suggest using that commit message instead of the PR title for the
          # merge commit, and it's easy to commit this by mistake. Enable this option
          # to also validate the commit message for one commit PRs.
          validateSingleCommit: true
```

https://github.com/amannn/action-semantic-pull-request/

I'm mobile right now or I'd PR the workflow yaml sorry :)

dulmandakh:
I made the change on GitHub, sorry 😇

PR #53, opened 2021-06-10
docs: mention the required imports AppDelegate.m

kuasha420:
This is not entirely obvious for non native developers.


mikehardy (approved):
A fair point - I believe it is covered in other related documents, but upstream, and without them the example is incomplete and would result in compile errors which is not a great developer experience. I like this change

Issue #21, opened 2021-04-19
Cannot run example app (crashes after infinite loop)


rahulsmehta:
## 🐛 Bug Report

When running the example app on an iOS simulator (in `example/ios`), the example app fails to mount & the app crashes after entering an infinite loop with the error:
```
Invariant Violation: No callback found with cbID 2903 and callID 1451 for module <unknown>.
```

## To Reproduce

1. Install both the package & example app dependencies (`npm install && npx pod-install`)
2. Start the example app by running `npm example:ios`

## Expected Behavior

The example app builds & runs successfully

## Environment

```
System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.87 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.16.3 - /usr/local/bin/node
    Yarn: 1.22.5 - ~/.yarn/bin/yarn
    npm: 7.10.0 - ~/git/sw/react-native-fbsdk-next/node_modules/.bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.6858069
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.9.1 - /usr/bin/javac
    Python: 3.7.7 - /opt/anaconda3/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1
    react-native: ^0.63.3 => 0.63.4
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

mikehardy:
I'm using the package in production now.
Not sure about the example though, I haven't tried it.

Luehang:
That's good to know.  What `react-native` and `react-native-fbsdk-next` version are you using?  And what platform?

mikehardy:
literally current stable every single thing as I type this - I maintain a pile of large repos so I keep everything up to date to catch the errors before my users do. Android and iOS

[...]

PR #68, opened 2021-07-06
Update index.js to import react-native-fbsdk-next

KrisLau:
The example is using `react-native-fbsdk` instead of `react-native-fbsdk-next` so I just thought I'd update that. It still doesn't run for me (see #21) but I thought I'd help out where I can for now as I do not know how to resolve the `InvariantViolation`

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -5,7 +5,7 @@
 import {AppRegistry, Platform} from 'react-native';
 import App from './src/App';
 import {name as appName} from './app.json';
-import {Settings} from 'react-native-fbsdk';
+import {Settings} from 'react-native-fbsdk-next';
 
 /**
  * The `autoInitEnabled` option is removed from facebook-ios-sdk, should initialize manually


mikehardy (approved):
nice catch

KrisLau:
Copying my comment over from #21 here as it might be pertinent:

> Ok so I found out more things but still can't fully resolve the error. For some reason named imports from 'react-native' aren't recognized so changing all the `import {a, b} from 'react-native'` to 
> ```
> import a from 'react-native'
> import b from 'react-native'
> ```
> 
> For example:
> ```
> import { requireNativeComponent, StyleSheet } from 'react-native';
> ```
> turns into
> ```
> import requireNativeComponent from 'react-native';
> import StyleSheet from 'react-native';
> ```
> 
> 
> This gets rid of a bunch of errors but leaves me with this error:
> ```
> [Tue Jul 06 2021 09:41:54.464]  BUNDLE  example/index.js 
> 
> [Tue Jul 06 2021 09:42:01.667]  ERROR    TypeError: undefined is not a function (near '..._reactNative.default.create...')
> [Tue Jul 06 2021 09:42:01.667]  ERROR    Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)
> [Tue Jul 06 2021 09:42:01.686]  ERROR    Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)
> ```
> which I believe is coming from `render(): JSX.Element;` in `FBLoginButton.ts` but I can't seem to figure our what to do for that

mikehardy:
Yeah when I saw the state of the example app, I wept.
Nothing frustrates me more than spending more than about 10 seconds on keeping an example app running.
I will post a PR in about an hour so with an auto-generation script for example apps a la https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh and/or https://github.com/mikehardy/rnfbdemo/blob/master/make-demo.sh

KrisLau:
> Yeah when I saw the state of the example app, I wept.
> Nothing frustrates me more than spending more than about 10 seconds on keeping an example app running.
> I will post a PR in about an hour so with an auto-generation script for example apps a la https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh and/or https://github.com/mikehardy/rnfbdemo/blob/master/make-demo.sh

@mikehardy Yeah I probably spent a little too long on it considering I couldn't get it to work anyway 😆 Thank you for everything you do to maintain this library btw, I really appreciate it!

mikehardy:
@KrisLau it exists - works for me on my machine ? #69

PR #69, opened 2021-07-06
docs(example): new example auto-generation script + results / docs changes to use it

mikehardy:
There is nothing more useless for module maintainers to do than try to keep up with react-native updates in their example apps.

But we're programmers, so we can just auto-generate an example via a script, then just re-run the script to update it.

I do this in a lot of other places to good effect:
- https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh
- https://github.com/mikehardy/rnfbdemo/blob/master/make-demo.sh

And now I propose it here.

**Each commit is separate so may be reviewed with a relevant commit message**, since the actual new example commit is huge (but also purely boilerplate / ignorable) as is the commit that kills the old example. Trying to look at the whole diff is a horrible thought, please do not torture yourself :-)

The test plan for this is to just keep re-running `refresh-example.sh` then `yarn example:ios` `yarn example:android` - and for me anyway, they work.

The previous example appeared broken

Supercedes / Closes #68
Related #21 

Interested party @KrisLau


.github/workflows/ci.yml
```diff
@@ -74,8 +74,8 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
       - name: Install dependencies
-        run: yarn install
+        run: yarn install && cd RNFBSDKExample && yarn
```
thebergamo, line 77: Why do we need to run yarn twice here?

README.md
```diff
@@ -183,15 +183,33 @@ Undefined symbols for architecture x86_64:
...
+    # Mixing Swift and Objective-C in a react-native project may be problematic.
+    # Workaround:  https://github.com/facebookarchive/react-native-fbsdk/issues/755#issuecomment-787488994
+    installer.aggregate_targets.first.user_project.native_targets.each do |target|
+      target.build_configurations.each do |config|
+        config.build_settings['LIBRARY_SEARCH_PATHS'] = ['$(inherited)', '$(SDKROOT)/usr/lib/swift']
+      end
+    end
+   ```
+
+  Both result in fixing search paths.
```
thebergamo, line 210: Good one!

thebergamo (approved):
Everything looks great! Thanks for this improved example :)

.github/workflows/ci.yml
mikehardy: One of them is at the top level, one of them is in the example. The example constructed here is less "magical" in that it does not automatically / auto-magically reference the parent directory code. That's a little inconvenient if using the example app for module development - and I could definitely see an improvement in the future that switches the new-style example to again reference the code raw on the filesystem from the parent directory, but in this style the example is a fully-standalone / not-integrated-with-anything react-native app. So that means you need to go into it and yarn install before it will run

mikehardy: (note, to implement this change - to have the example again reference the parent directory module code, I think it's just a switch in build.gradle/Podfile/metro.config.js to put the references in place, or alternatively perhaps it can just be an override of the `yarn add` command to be `yarn add ../` or something? But then it will still be referencing *just the module*, not the rest of the modules and thus it will still need a yarn run in the example itself

README.md
mikehardy: Apple Silicon is still *so* difficult to handle, it's crazy

PR #75, opened 2021-07-15
Upgrade Facebook SDK to latest version 11.0 

lucaspelloni2:


mikehardy:
How'd you get it to compile against the new SDKs with their breaking changes, by only changing the podspec :thinking:  ?

PR #82, opened 2021-07-29
Refactor-example-podlock-file

passerbyloo:
example ios project run `pod install` will popup error.

```
[!] CocoaPods could not find compatible versions for pod "FBSDKLoginKit":
  In snapshot (Podfile.lock):
    FBSDKLoginKit (= 9.0.1, ~> 9.0.1)

  In Podfile:
    react-native-fbsdk-next (from `../../`) was resolved to 4.3.0, which depends on
      react-native-fbsdk-next/Login (= 4.3.0) was resolved to 4.3.0, which depends on
        FBSDKLoginKit (~> 9.3)

```

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -9,20 +9,20 @@ PODS:
     - React-Core (= 0.63.4)
     - React-jsi (= 0.63.4)
     - ReactCommon/turbomodule/core (= 0.63.4)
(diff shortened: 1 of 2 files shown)


mikehardy:
Thank you for taking the time and proposing this - however I think #69 designs out the need for this completely by auto-generating the example so there is no manual effort required ever again

At least, that was the goal!

PR #89, opened 2021-08-23
fix(android, event-emitter): add required event listener stubs for react-native 0.65

mikehardy:
These are needed to avoid console warnings about NativeEventEmitter being
created without the required native module implementations

Test Plan:

It's impossible for these methods to have bugs because they're just stubs ;-)

I've integrated it locally via patch-package, and it's the same change I've successfully made + released in react-native-firebase and react-native-device-info, as well as patched locally in netinfo (PR there is merged but not released yet).


mikehardy:
https://github.com/invertase/react-native-firebase/pull/5616
https://github.com/react-native-device-info/react-native-device-info/pull/1289
https://github.com/react-native-netinfo/react-native-netinfo/pull/487

PR #95, opened 2021-09-12
refactor: Migrate to typescript

g-dimitry:
Hello Guys,

I've successfully migrated all the code base to Typescript. My philosophy was to try as much as I can to make this a minor release (as @mikehardy stated). You'll see I have added three TODO comments (search for "TODO: [TS Migration]:"). Those are changes that need to be discussed and addressed maybe in the next major release.

To test the compatibility of both the new types and the old types:

1. I reverted the commit removing the old types
2. Created an index.ts file at the root of the repo
3. Imported both new and old types the same way so: `import * as oldTypes from 'react-native-fbsdk-next'` and `import * as newTypes from './src'`
4. Created a new constant with the same type as the oldTypes' import and assign to it the newTypes: `const o: typeof oldTypes = newTypes;`
5. Solved all differences until the compiler had no errors. (All changes are marked by a "TODO: [TS Migration]:" comment)

I guess we need to test it on a running project first to check it works with a JS project & a TS project in development and release modes.

If anybody can test the above and confirm it's not breaking it would be awesome.

Looking forward to your comments.


g-dimitry:
And of course the eslint/prettier configurations should be changed.

g-dimitry:
I just changed the Eslint configuration but I had to change A LOT of code that may break old code so I guess the best way to do this is to agree on what to do with this migration: should we release a minor release and then a new major release with the linted and clean code (that may break some old false types) ?

mikehardy:
1) this sounds amazing!
2) with sincere apologies I'm traveling until the 21st and then will likely have a backlog of project work to slash through before I can spin through volunteer repos so perhaps someone else can give this a spin 🙏  ? Otherwise it'll have to sit for a little bit ...

mikehardy:
(I can say philosophically just in my opinion I think a major is probably warranted, the goal was just for it to be easiest possible migration for any module users, as a general courtesy - I'm guessing it's probably walking that fine line well just based on how you're describing it though I admit I haven't looked)

thebergamo:
I was considering picking up this task together with upgrading Facebook SDK.
Can't thank you more @g-dimitry for pick this challenge of migration to TS.
Let me review it o/

src/FBAppEventsLogger.ts
```diff
@@ -131,7 +133,7 @@ module.exports = {
   logEvent(eventName: string, ...args: Array<number | Params>) {
     let valueToSum = 0;
     if (typeof args[0] === 'number') {
-      valueToSum = args.shift();
+      valueToSum = args.shift() as number;
```
thebergamo, line 136: Maybe instead of using ’as’ here, you could use ’Number(args.shift()’
What do you think?

src/FBGraphRequestManager.ts
```diff
@@ -49,12 +47,14 @@ function _verifyParameters(request: GraphRequest) {
...
   requestBatch: Array<GraphRequest>;
-  requestCallbacks: Array<?Callback>;
+  requestCallbacks: Array<Callback | undefined>;
   batchCallback: Callback;
 
   constructor() {
     this.requestBatch = [];
     this.requestCallbacks = [];
+    // TODO: [TS Migration]: Added default value to fix type definitions. Discuss before merging.
+    this.batchCallback = () => {};
```
thebergamo, line 57: Isn't the case here to just use `this.batchCallback?.()` and then you won't need to put a default value here?
But either way having a default noop here seems fine to me.

src/FBLoginButton.tsx
```diff
@@ -17,23 +17,21 @@
...
+import {
   DefaultAudience,
   LoginBehaviorAndroid,
   LoginResult,
   LoginTracking,
 } from './FBLoginManager';
 
-type Event = Object;
-type TooltipBehaviorIOS = 'auto' | 'force_display' | 'disable';
+export type Event = any;
```
thebergamo, line 33: Can't we be a bit more strict here?

src/FBProfile.ts
```diff
@@ -73,17 +70,18 @@ class FBProfile {
     this.linkURL = profileMap.linkURL;
     this.imageURL = profileMap.imageURL;
     this.userID = profileMap.userID;
-    this.email = Platform.OS === 'android' ? null : profileMap.email;
+    // TODO: [TS Migration]: Value should be undefined. Discuss before merging.
+    this.email = Platform.OS === 'android' ? null as unknown as undefined : profileMap.email;
```
thebergamo, line 74: In this case we could just mark `email: string? | null` and we can avoid this unknown type

src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import FBLoginManager from './FBLoginManager';
+import FBMessageDialog from './FBMessageDialog';
+import FBProfile from './FBProfile';
+import FBSettings from './FBSettings';
+import FBShareDialog from './FBShareDialog';
+import FBLoginButton from './FBLoginButton';
+import FBSendButton from './FBSendButton';
+import FBShareButton from './FBShareButton';
+
+// TODO: [TS Migration]: Removed getters and used normal export. Discuss before merging.
```
thebergamo, line 42: Seems fine for me as it should have similar manner while importing

thebergamo (changes requested):
There are few questions of mine also inside some points, specially this `name: string | undefined` as I believe we should stick to a pattern in this case using `name?: string` as in most cases would have similar effect.

Once again thanks for this PR o/

src/FBProfile.ts
g-dimitry: This will break

g-dimitry:
@thebergamo I'll finalize this on Saturday and I'll be sure to specify all breaking changes and tag you to discuss before merging. Thank you for the collaboration invitation, I'm very excited to help 😄

thebergamo:
@g-dimitry i guess you would need to rebase as some new feature is added :)

g-dimitry:
@thebergamo Unfortunately yes, Sorry for being late, Fell horribly ill. I guess I'll merge master into my branch it would be easier as I have 18 commits which will be a nightmare to rebase.

thebergamo:
That's ok mate, your health is much more important o/ get great recovery 😃

rnike:
Just saw this before making the new implementation, does this need help? Maybe I can help complete this before implementing the new feature.

mikehardy:
@mike this one really needs help - yes! It would be fantastic to have this be in typescript and I don't believe the library has moved on that far since it was initially begun here

rnike:
okay let me take a look at it

rnike:
Commits in this pr have been merged in #160, thanks for all the works.

PR #97, opened 2021-09-20
fix: NSStringArray warnings

passerbyloo:
fix Xcode build warnings: `'NSStringArray' is deprecated: Use NSArray<NSString *>`
sync types with [facebook-ios-sdk](https://github.com/facebook/facebook-ios-sdk/blob/main/FBSDKCoreKit/FBSDKCoreKit/FBSDKSettings.m#L433)


mikehardy (approved):
These look good to me, thanks @passerbyloo !
I checked the code in facebook-ios-sdk and it looks like they've had these types the whole time, FWIW, via examination of git blame

passerbyloo:
thanks for your verification

PR #99, opened 2021-09-21
feat: add logProductItem

santiagofm:
This function has been on both android & ios native SDKs for a couple of years now... just exposing it to RN land

**Test plan**
used this branch on our app and logged some events, validated they show up on Facebook's Events Manager debugger view:

iOS
<p align="center">

</p>
Android
<p align="center">

</p>

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
@@ -38,6 +38,7 @@ import com.facebook.react.module.annotations.ReactModule;
 import java.math.BigDecimal;
 import java.util.Currency;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -141,7 +142,7 @@ public class FBAppEventsLoggerModule extends ReactContextBaseJavaModule {
      */
     @ReactMethod
     public void setFlushBehavior(String flushBehavior) {
-        AppEventsLogger.setFlushBehavior(AppEventsLogger.FlushBehavior.valueOf(flushBehavior.toUpperCase()));
+        AppEventsLogger.setFlushBehavior(AppEventsLogger.FlushBehavior.valueOf(flushBehavior.toUpperCase(Locale.ROOT)));
     }
 
     /**
@@ -232,8 +233,8 @@ public class FBAppEventsLoggerModule extends ReactContextBaseJavaModule {
                                String brand,
                                Bundle parameters) {
         mAppEventLogger.logProductItem(itemID,
-                AppEventsLogger.ProductAvailability.valueOf(availability.toUpperCase()),
-                AppEventsLogger.ProductCondition.valueOf(condition.toUpperCase()),
+                AppEventsLogger.ProductAvailability.valueOf(availability.toUpperCase(Locale.ROOT)),
+                AppEventsLogger.ProductCondition.valueOf(condition.toUpperCase(Locale.ROOT)),
                 description,
                 imageLink,
                 link,
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
@@ -20,6 +20,8 @@
 
 package com.facebook.reactnative.androidsdk;
 
+import android.os.Bundle;
+
 import androidx.annotation.Nullable;
 
 import com.facebook.appevents.AppEventsConstants;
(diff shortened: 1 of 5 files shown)


dmedina-vopero (approved):

icattivelli (approved):

jdalto (approved):

mgiglio (approved):

patsanes (approved):

android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
```diff
@@ -198,6 +200,52 @@ public void logPushNotificationOpen(@Nullable ReadableMap payload) {
...
+                               String title,
+                               BigDecimal priceAmount,
+                               Currency currency,
+                               String gtin,
+                               String mpn,
+                               String brand,
+                               Bundle parameters) {
+        mAppEventLogger.logProductItem(itemID,
+                AppEventsLogger.ProductAvailability.valueOf(availability.toUpperCase()),
+                AppEventsLogger.ProductCondition.valueOf(condition.toUpperCase()),,
```
mikehardy, line 236: Nice! Thanks for posting this!
I worry a little about these lines for two reasons:

1) if you call String.toUpperCase without a null check, I'm sure there will be crashes with null pointer exceptions later. Should the javascript enforce that these elements are non-nonnull prior to sending them down to native maybe? A follow-on thought is that all of the items may have requirements from the facebook SDK and I didn't see that enforced anywhere other than the types (which are optional in javascript - people may accidentally send in whatever value or undefined...)

2) if you call any of the string mutation operations like toUpperCase without specifying a locale (I suggest `Locale.ROOT` is probably fine?) you can end up with *really* hard to fix errors like this fiendish one https://github.com/invertase/react-native-firebase/issues/3917#issuecomment-843401492

mikehardy (changes requested):
As mentioned in a line-specific comment, I think this is awesome, but I'm curious about nullability / validation of arguments and the use of toLowerCase without Locale

android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
santiagofm: Thanks for the quick reply @mikehardy! Heres what i think:

1. Both native SDKs already provide checks for required parameters [here](https://github.com/facebook/facebook-android-sdk/blob/main/facebook-core/src/main/java/com/facebook/appevents/AppEventsLoggerImpl.kt#L198-L231) and [here](https://github.com/facebook/facebook-ios-sdk/blob/989061abaf9e5ec35e7d3d76b38073fa6f60c269/FBSDKCoreKit/FBSDKCoreKit/AppEvents/FBSDKAppEvents.m#L568-L595) so I think we are covered on that end. I only added a null check for those 3 trailing parameters because at least one of them is required, but tbh I'm not sure if throwing an error is even the correct way of handling this case, what do you think?
As for passing in some other type than the one we expect, oh well I guess that's just inherent to using javascript and there's so much we can do. If you really think this type of check would be needed please point me towards the right way of doing it and i'll gladly do it :)

2. Sure thing! I'll _fix_ other usages of the same function too.

mikehardy: Hey @santiagofm - re: your reply on point 1, the native SDKs do null check + validate (thanks for the link) so I could agree we are mostly covered there, but if you call toLowerCase on a null string  you will null pointer exception and crash prior to ever hitting the native SDK, so that still needs to be addressed

I agree it's standard javascript that people can send anything, but libraries seem to take two approaches here. Some (like react-native-firebase which I work on) actually validate everything and have a little set of javascript parameter validator utils, and others just trust the developer. fbsdk seems to just trust the developer so I'm fine with continuing to not really validate, but null pointer exceptions must be eliminated as a possibility I think or we will certainly see an issue here later and have to do it anyway

Thanks for throwing the Locale.ROOT in there, I think in general that is important but those errors pop up in very unexpected ways, so nice to eliminate them as a class

mikehardy (changes requested):
As mentioned in general comment, still vulnerable to a null pointer exception prior to going down to native SDK where they are checking/validating - just needs a null check for those two items prior to calling toUpperCase

android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java
santiagofm: That's great! I'll source those validation helpers from react-native-firebase !

mikehardy: Okay - if you want to do that I think they are all in here and something similar could work https://github.com/invertase/react-native-firebase/blob/master/packages/app/lib/common/validate.js - but as mentioned above I'm not sure it's necessary, I'm most concerned with just making sure we don't actually crash on a NullPointerException in those two specific lines of the java imlementation where toUpperCase are used on possibly null input

mikehardy (approved):
Oh wow, that takes validation all the way, very nice. I can't see how this could crash now or in the future, thank you!
Final phase for me is just to verify that you have in fact run this current version locally (via patch-package or similar) and it's all working for you?

santiagofm:
> Oh wow, that takes validation all the way, very nice. I can't see how this could crash now or in the future, thank you!
> Final phase for me is just to verify that you have in fact run this current version locally (via patch-package or similar) and it's all working for you?

Thanks for the question because I'd forgotten to test on Android and it wouldn't work, I was using some non-primitive types for the parameters that aren't supported by the RN bridge. That is sorted out on the latest commit and I added some screenshots from Facebook's event manager debugger view which I used to validate this is working on both platforms. 🎉

mikehardy (approved):
nice - normally I would have more fears about null / empty arguments then going through processing on Android (the new BigDecimal.valueOf and Currency.getInstance) but since you validate them all in javascript already (I checked those ones) this still looks great

santiagofm:
Hi @thebergamo! Could you review this PR? Thanks!

thebergamo:
It indeed looks very nice :)

santiagofm:
> It indeed looks very nice :)

Awesome, thanks!

Could we get a new release? :) cc @mikehardy

PR #101, opened 2021-09-28
chore(test): add testID optional type to login button for use in testing

doublethefish:
Test Plan:
The commit itself adds a test to demonstrate a usecase.

### DIFF
--- a/RNFBSDKExample/App.js
+++ b/RNFBSDKExample/App.js
@@ -57,6 +57,7 @@ export default class App extends Component<{}> {
     return (
       <View style={styles.container}>
         <LoginButton
+          testID="facebook-login"
           onLoginFinished={(error, data) => {
             Alert.alert(JSON.stringify(error || data, null, 2));
           }}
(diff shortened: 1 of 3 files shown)


doublethefish:
I am not sure why the `PR Title is Conventional` check fails with:

> No release type found in pull request title "chore(test) Adds type-hint for testID for downstream testing". Add a prefix to indicate what kind of release this pull request corresponds to

Happy to change it to something that passes.

mikehardy:
conventional commits mandates a colon separating the type and the message - that's the issue I think https://www.conventionalcommits.org/en/v1.0.0/

mikehardy (approved):
Cool - this change looks obviously good to me, I'll change the PR title but I think the checker will still fail (mostly as a warning now, for maintainers) since GitHub will by default take the commit message on single commits, but if I squash merge I can make sure it's right (in other words: no action required to make it conventional)

mikehardy:
CI passed, and I just re-read it again to make sure it still looked good to me just in case, I think this is good to go, thanks again!

mikehardy:
Note that I actually merged it as a feature - when I was considering it, this is actually a new feature for the library (even if it is only for use in testing downstream) and it won't generate a release of the right type unless committed as such.

doublethefish:
Grand, thank you.

PR #103, opened 2021-09-29
fix(type): add new testID to typescript definition file

doublethefish:
Adds `testID` to the typescript type information.

Test Plan:
Testing the module in a production app compiled with tsc.

### DIFF
--- a/types/FBLoginButton.d.ts
+++ b/types/FBLoginButton.d.ts
@@ -57,6 +57,11 @@ export class LoginButton extends React.Component<{
      * View style, if any.
      */
   style?: any;
+
+  /**
+   * testID, if any.
+   */
+  testID?: string,
 }> {
   static defaultProps: {
     style: typeof styles.defaultButtonStyle;
--- a/types/FBLoginButton.d.ts
+++ b/types/FBLoginButton.d.ts
@@ -61,7 +61,7 @@ export class LoginButton extends React.Component<{
   /**
    * testID, if any.
    */
-  testID?: string,
+  testID?: string;
 }> {
   static defaultProps: {
     style: typeof styles.defaultButtonStyle;


types/FBLoginButton.d.ts
```diff
@@ -57,6 +57,11 @@ export class LoginButton extends React.Component<{
      * View style, if any.
      */
   style?: any;
+
+  /**
+   * testID, if any.
+   */
+  testID?: string,
```
mikehardy, line 64: Oh of course, I should have seen that yesterday when I reviewed your implementation PR. Thanks for the follow-up

I think it needs to be like this though? It looks like we are inconsistent on this point actually! Most are semicolon-terminated (correct I think) while some are commas. I wonder if our types are being tested correctly :thinking: . Do you confirm this works for you? Can you confirm it works with the semi-colon (which I think is more correct)?

```suggestion
  testID?: string;
```

mikehardy (changes requested):
tiny question on semi-colon vs comma termination on typedef

mikehardy:
Ah, this won't be caught until #95 is done most likely. The types are not checked anywhere.

types/FBLoginButton.d.ts
doublethefish: Good catch, thank you. 

I had tested it locally with the correct fix but somehow submitted this... hmmm.

mikehardy (approved):
LGTM thanks!

mikehardy:
Not sure what is going on with commas vs semi-colons but I discovered in my work project I was actually nulling out the types here anyway, from before there were types here. I just started using the internal types after updating to the new 4.5.0 release that has this change among others, and everything seems fine right now, so this is working for me :-)

PR #125, opened 2021-11-08
feat(ios): Upgrade to Facebook iOS SDK 11.2.1

ArsenyYankovsky:
This PR upgrades [Facebook SDK for iOS](https://github.com/facebook/facebook-ios-sdk) to 11.2.1.
Method `updateUserProperties` of the `RCTFBSDKAppEvents` class was removed in release 11.0.0 ([changelog](https://github.com/facebook/facebook-ios-sdk/blob/main/CHANGELOG.md#1100)).

Test Plan:

My app mostly logs events so I tested that it compiles, launches and events are visible in the "Test Events" tab in the Facebook Events Manager.

### DIFF
--- a/ios/RCTFBSDK/core/RCTFBSDKAppEvents.m
+++ b/ios/RCTFBSDK/core/RCTFBSDKAppEvents.m
@@ -149,13 +149,6 @@ RCT_EXPORT_METHOD(getAdvertiserID:(RCTPromiseResolveBlock)resolve
   }
 }
 
-RCT_EXPORT_METHOD(updateUserProperties:(NSDictionary *)parameters)
-{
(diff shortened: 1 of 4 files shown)


mikehardy (approved):
This looks fantastic, to me, thank you very much for doing this!
After a quick scan of the changelogs it seems like doing Android to v11 might not be so bad, and to v12 on android also not so bad, but the v12 jump for iOS looks pretty big unfortunately. Any thoughts on those?

mikehardy:
I've done the matching change to the android sdk and removed the same method (regarding user properties) from the entire library since it's removed there, so it's all one breaking change with one version bump - we're close to current on native SDKs now! Just iOS v12 pending

PR #137, opened 2021-11-17
fix: Profile.getCurrentProfile() always returns null in the first time

### LINKED ISSUE #73: [Android] Profile.getCurrentProfile() always returns null in the first time
Can you tracking profile and return data instead of return null ?
This is my solution: change some lines in android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java 

replace
```java   
 @ReactMethod
    public void getCurrentProfile(Callback callback) {
        //Return the profile object as a ReactMap.
        callback.invoke(Profile.getCurrentProfile() == null
                ? null
                : Utility.profileToReactMap(Profile.getCurrentProfile()));
  }
```
with something like this:
```java
   import com.facebook.ProfileTracker;
   ...
   private ProfileTracker mProfileTracker;
   
   @ReactMethod
   public void getCurrentProfile(final Callback callback) {
        //Return the profile object as a ReactMap.
         if(Profile.getCurrentProfile() == null) {
                mProfileTracker = new ProfileTracker() {
                    @Override
                    protected void onCurrentProfileChanged(Profile oldProfile, Profile currentProfile) {
                        callback.invoke(Utility.profileToReactMap(currentProfile));
                        mProfileTracker.stopTracking();
                    }
                };
            }
             else {
                 callback.invoke(Utility.profileToReactMap(Profile.getCurrentProfile()));
                }

  }
```

SMKH-PRO:
fix: "Profile.getCurrentProfile() always returns null in the … #137

fixes #73

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
@@ -1,5 +1,5 @@
 package com.facebook.reactnative.androidsdk;
-
+import com.facebook.ProfileTracker;
 import com.facebook.Profile;
 import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.NativeModule;
@@ -16,6 +16,7 @@ import androidx.annotation.NonNull;
 @ReactModule(name = FBProfileModule.NAME)
 public class FBProfileModule extends ReactContextBaseJavaModule {
     public static final String NAME = "FBProfile";
+    private ProfileTracker mProfileTracker;
 
     public FBProfileModule(ReactApplicationContext reactContext) {
       super(reactContext);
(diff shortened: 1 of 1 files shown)


mikehardy:
Great! Thanks

Did you see my comments on the related issue?

> Please pay special attention to the documentation for it, to make sure people understand it is a Promise (asynchronous resolution) and could be a bit delayed.

I verified the doc has example code with a ".then", so this should be okay

> I would also want to make sure that it timed out in some way - is there a chance this could hang and never resolve? That would not be acceptable - it would violate the caller's expectation of the API, so if there is not a timeout in the underlying call we would need to make it interruptible and set a timeout ourselves

This is potentially worrisome. It may be rare but what if the network goes out in between login and getProfile, this should eventually timeout I think? Will it? Can you try making this call like so: 1) login, 2) disable your network 3) then try to access this method. What happens? Does it work?

thebergamo:
@SMKH-PRO any chance to that you have addressed those points from @mikehardy ?

yalcinozer:
Unfortunately I'm having this issue also. I have stable internet. I'm waiting until promise resolved. On first try, it returns null however it returns user object without problem in all next tries. It is the same code block actually.

PR #140, opened 2021-11-19
feat: jest tests

diegotsi:
Hey guys! Recently I needed to create some mocks for this lib, so I'm bringing this to the lib to help others that want to do the same.

These are the features from this PR:

1 - Configure Jest
2 - Create a mock folder with `AppEventParams` and `AppEvents` to make a more faithful mock, as this content came from the Native we can't access this with `AppEventsLogger.AppEvents` on Jest, that's why I'm putting this on the mocks folder.
3 - Update readme with a basic example module.

Screen Shot 2021-11-19 at 10 52 05


package.json
```diff
@@ -62,7 +62,8 @@
     "example:clean": "cd RNFBSDKExample && \\rm -fr yarn.lock node_modules ios/Podfile.lock && cd ..",
     "example:install": "cd RNFBSDKExample && yarn && cd ios && (pod install || true) && cd ../..",
     "example:devcopy": "yarn prepare && cp -rv android ios lib src types *.podspec RNFBSDKExample/node_modules/react-native-fbsdk-next/",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release",
+    "jest": "jest"
```
mikehardy, line 66: Very interesting, how does CI go when you change package.json line 55 to include ` && yarn jest` so this is exercised?

jest/setup.js
```diff
@@ -0,0 +1,23 @@
...
+    },
+    LoginManager: {
+      logInWithPermissions: jest.fn(),
+    },
+    Settings: {
+      initializeSDK: jest.fn(),
+    },
+    LoginButton: nonMocked.LoginButton,
+  };
+});
```
mikehardy, line 23: could these all be in the mock (and could the mock export the real symbol names) so that the module is fully tested by library users, with just one line?
I thought the react-native-permissions usage (one line in setup) was super clean, for instance:
https://github.com/zoontek/react-native-permissions#testing-with-jest

.eslintrc.js
```diff
@@ -23,4 +23,9 @@
 
 module.exports = {
   extends: '@react-native-community',
+  env: {
+    jest: true,
```
mikehardy, line 27: this is interesting - I have not had to add a jest env to my eslints before, what happens if this is not included?
The react-native template just does a react-native template does not do this https://github.com/facebook/react-native/blob/765844055ba0d02262a11114bad5da67e935d8df/template/_eslintrc.js#L3 :thinking: - I am probably just missing something as this is new to me

It appears it is pulling this plugin https://github.com/jest-community/eslint-plugin-jest#readme and that is to enforce good jest practices? 

But the plugin is already there from `@react-native-community` ? https://github.com/facebook/react-native/blob/765844055ba0d02262a11114bad5da67e935d8df/packages/eslint-config-react-native-community/index.js#L27

So I am confused :-)

mikehardy (changes requested):
This is really cool! Such a common need. I had a few thoughts on various parts and left comments - curious what you think

package.json
diegotsi: Hey @mikehardy, I think we can put this on line 55 to run with the test script, I'll try it.

jest/setup.js
diegotsi: Yes! I'll check it

.eslintrc.js
diegotsi: This is to avoid this error


Screen Shot 2021-11-19 at 11 30 06


We need to tell eslint that we are using jest.

package.json
diegotsi: it's working 😄

jest/setup.js
diegotsi: That was an awesome tip, it's more cleaner now!

package.json
diegotsi: 2fe289a7d2ec3d5bc9583fb573574dd0b0c740fc

.eslintrc.js
mikehardy: I get that - I am really trying to figure out why we extend `@react-native-community` and in there *it does this*

https://github.com/facebook/react-native/blob/765844055ba0d02262a11114bad5da67e935d8df/packages/eslint-config-react-native-community/index.js#L66-L69

It should already be configured. This is very confusing as it appears we're just doing what we have done already. Why isn't `@react-native-community` working ? If the jest part is not working, is there some other part not working ?

package.json
mikehardy: fantastic!

.eslintrc.js
diegotsi: I'll check it @mikehardy

README.md
```diff
@@ -661,6 +661,24 @@ yarn example:android
 Note: You'll probably want to change the Facebook App ID to your own, else the example app won't be able to login.  To change it, edit your local copy of `refresh-example.sh`, update the `FacebookAppId` variable, then re-run `refresh-example.sh` to regenerate the example directory.
 
 
+## Testing with Jest
+
+We have a example mock inside `jest/setup.js` but you just add the following line to your setup file:
+
+```js
+jest.mock('react-native-fbsdk-next', () => require('react-native-fbsdk-next/jest/mocks').default);
```
mikehardy, line 669: I don't think this will be present in the published package unless this array includes the jest directory ?

https://github.com/thebergamo/react-native-fbsdk-next/blob/e25a711d2d1b3557733cf98ed15ff7ed6b8eecde/package.json#L67-L78

jest/setup.js
```diff
@@ -0,0 +1 @@
+jest.mock('../src', () => require('./mocks').default);
```
mikehardy, line 1: that's a really clean integration experience! nice

mikehardy (changes requested):
The CI addition is fantastic, and the single jest setup line is fantastic
I left a note on including it in the released module, that's important :-)
I'm still confused about the eslint integration - I just don't understand that. We don't have to block the PR on it but if I can't understand it I would at least like a comment in there that has the URL to this discussion and a note "this may not be necessary?"

So the package.json change is the blocker, the eslint thing I'd like to understand but could let it go

.eslintrc.js
diegotsi: @mikehardy i think i understand it, look the override from `react-native-community`:

```
files: [
  '*.{spec,test}.{js,ts,tsx}',
  '**/__{mocks,tests}__/**/*.{js,ts,tsx}',
],
env: {
   jest: true,
  'jest/globals': true,
},
rules: {
  'react-native/no-inline-styles': 0,
  quotes: [1, 'single', {avoidEscape: true, allowTemplateLiterals: true}],
},
}, 
```

they're applying override only for files that contains `spec` or `test`, that's why is not working for us, because my setup file is just a `index.js` file.

What do you think?

README.md
diegotsi: done 517860910612865d4f4ea07a2ce1e806fff5cca6

.eslintrc.js
mikehardy: perhaps reduce complexity by having it be index.test.js ? would that work - fewer moving parts / smaller changes are usually a plus

diegotsi: My opinion about this is that a file with `test`  or `spec` should have a testing case, the setup file should not use one of those prefixes.

That's how I see it, what about you?

mikehardy: My opinion of putting it in the env with no filter is that it defines jest globals in files that should not have them. The setup file is definitely test related. Perhaps as a middle ground, just define the jest global directly in the mock using the eslint syntax to do so ? 

> /* eslint-env jest */

This is a super super minor point though of course, I don't want the general awesomeness of having jest integrated in CI and a mock available to everyone to be overshadowed at all!

diegotsi: that's an awesome idea! done 278a337c7d57618a42225246823d3403dba7cac1

mikehardy (approved):
This is a thing of beauty, thank you!

https://github.com/thebergamo/react-native-fbsdk-next/runs/4269445913?check_suite_focus=true#step:8:10
```
$ jest
PASS RNFBSDKExample/__tests__/App-test.js
  ✓ renders correctly (717 ms)
```

PR #144, opened 2021-11-22
fix: adjust import and jest eslint env

diegotsi:
Hello! After this [merge](https://github.com/thebergamo/react-native-fbsdk-next/pull/140) I tried to update the lib and use the mocks but I'm getting the following error:

Screen Shot 2021-11-22 at 11 04 21

So I change the import to the component itself like react-native-permission [does](https://github.com/zoontek/react-native-permissions/blob/master/mock.js) and started working on the app.

Also, I'm adding the `/* eslint-env jest */` to avoid eslint error on this file (didn't see this on the first PR)

### DIFF
--- a/jest/mocks/index.js
+++ b/jest/mocks/index.js
@@ -1,4 +1,5 @@
-const nonMocked = require('../../lib/commonjs');
+/* eslint-env jest */
+const nonMocked = require('../../src');
 
 export const mockAppEvents = {
   AchievedLevel: 'fb_mobile_level_achieved',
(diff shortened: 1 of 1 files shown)


diegotsi:
@mikehardy can you check this?

jest/mocks/index.js
```diff
@@ -1,4 +1,6 @@
-const nonMocked = require('../../lib/commonjs');
+/* eslint-env jest */
```
mikehardy, line 1: Oh dang! We got it one spot https://github.com/thebergamo/react-native-fbsdk-next/pull/140/files#diff-759a02c8fa99d555891bb2a2d3ae2495a2295dc1e1183dc8ba6281b0e172208aR1

but missed this one. Nice catch

mikehardy (approved):
If this works for you and works in CI (which tests jest now...) that works for me
Thanks for taking the time to help everyone with this fix!

PR #154, opened 2021-12-16
feat(ios, sdk): facebook-ios-sdk from 12.1.0 to 12.2.1

mikehardy:
Update to latest underlying SDK - refresh example in the process of local testing

https://github.com/facebook/facebook-ios-sdk/blob/main/CHANGELOG.md#1221

### DIFF
--- a/RNFBSDKExample/ios/Podfile.lock
+++ b/RNFBSDKExample/ios/Podfile.lock
@@ -2,8 +2,8 @@ PODS:
   - boost (1.76.0)
   - CocoaAsyncSocket (7.6.5)
   - DoubleConversion (1.1.6)
-  - FBAEMKit (12.1.0):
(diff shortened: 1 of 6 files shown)


mikehardy:
updated the example so it compiles correctly with it (podfile.lock wasn't happy with it when testing locally)
this seems to work locally, I can't think of a reason not to push it live if CI goes green

thebergamo:
Lgtm. Can't we automate these pull  requests just to update the internal libs when just minors/patches ?

thebergamo (approved):

mikehardy:
With enough time and desire yes :sweat_smile: - it's basically a dependabot function, but they don't support pods at the moment https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem

PR #155, opened 2021-12-16
refactor!: migrate to typescript

rnike:
continue #95

Resolve todos and fix some issues, ~~also changed the compiler to `tsc` because this is how I usually do, `@react-native-community/bob` was removed.~~

This is a huge change, before public release, I think we should have a beta version first


rnike:
the conflict has been resolved

mikehardy:
Ironically I just spent most of my "open source" time today reviewing another module's typescript re-write, so I don't have time just this moment - but similar to that one I cannot overstate how happy I am to see this, and how happy I will be to have it in typescript. So thank you

One quick question - the switch from bob to tsc. using bob let's us easily generate mutliple derivative JS targets, commonjs and modules, at the same time. So is kind of useful. Do you feel strongly about raw tsc, or even if bob is unfamiliar to you, would it be okay to leave it and preserve our ability to intermediate between our typescript source and various/possibly-multiple JS transpiles?

rnike:
@mikehardy The main point I remove bob is I didn't see why we need multiple kinds of js output, and I kind of want to reduce the package size on npm (including not uploading the src folder) to just make it neat, but maybe the whole thing (bob removal and size reduce) should be fully discussed before I go further, I'm a little bit too fast on this without consideration, let me put bob back on this PR, thanks for the reminder.

src/FBGraphRequestManager.ts
```diff
@@ -17,24 +17,26 @@
...
 function _verifyParameters(request: GraphRequest) {
-  if (request.config && request.config.parameters) {
-    for (var key in request.config.parameters) {
-      var param = request.config.parameters[key];
-      if (typeof param === 'object' && param.string) {
+  if (request.config?.parameters) {
+    for (const key in request.config.parameters) {
+      const param = request.config.parameters[key];
+
+      if (typeof param === 'object' && (param as any)?.string) {
```
rnike, line 39: `(param as any)` causes the warning, but I currently cannot figure out the solution

mikehardy:
Just about including source or not: uploading src is a strict requirement for me. The majority of pull requests come from people that just reach into node_modules and just fix something, the PR is an afterthought. If you remove the src and build machinery you lose a huge percentage of contributions in my experience. I do not see uploading a few extra text files in the npmjs pack binary as being any sort of problem, compared to the general problem that in open source, people that actually contribute are far too infrequent...

rnike:
@mikehardy agree with you uploading a few extra text files isn't a big deal, I've added bob back and this PR I think is ready to merge if no further issue.

mikehardy:
great - it's the weekend here in the holidays so review might be a bit delayed but please don't take that as any signal - I'm really excited to get this done. Thanks again for posting it

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Facebook SDK support for React Native apps.",
   "main": "lib/commonjs/index.js",
   "module": "lib/module/index.js",
-  "react-native": "src/index.js",
+  "types": "lib/typescript/index.d.ts",
+  "react-native": "src/index.ts",
```
rnike, line 8: Probably will lead to an issue, not everyone is using typescript, I should remove this

rnike:
It will be wonderful if there's a beta version on npm to test it properly, currently I am just pushing it to my repo and yarn add from there.

None is it necessary, just consider it as a `nice to have`

src/FBAccessToken.ts
```diff
@@ -148,7 +146,7 @@ class FBAccessToken {
    * listener when called.
    */
   static addListener(
-    listener: (accessToken: ?FBAccessToken) => void,
+    listener: (accessToken: FBAccessToken | null) => void,
```
rnike, line 149: check should it be `accessToken?: FBAccessToken`

src/FBAppEventsLogger.ts
```diff
@@ -68,87 +67,87 @@ type ProductAvailability =
...
 } = AppEventsLogger.getConstants();
 
-module.exports = {
+export default {
   /**
    * Sets the current event flushing behavior specifying when events
    * are sent back to Facebook servers.
    */
-  setFlushBehavior(flushBehavior: AppEventsFlushBehavior) {
+  setFlushBehavior(flushBehavior: AppEventsFlushBehavior): void {
```
rnike, line 150: not sure `: void` is necessary, remove it if is not

rnike: unnecessary type declarations have been removed

src/FBAccessToken.ts
rnike: it is surely `FBAccessToken | null` according to the native implementation

src/FBGraphRequest.ts
```diff
@@ -56,24 +56,24 @@ class FBGraphRequest {
...
    * The optional config for the request.
    */
-  config: ?GraphRequestConfig;
+  config: GraphRequestConfig | undefined;
 
   /**
    * Called upon completion or failure of the request.
    */
-  callback: ?GraphRequestCallback;
+  callback: GraphRequestCallback | undefined;
```
rnike, line 64: according to the constructor, `callback` will never be undefined,  so does `config`

src/models/FBShareOpenGraphValueContainer.ts
```diff
@@ -17,28 +17,29 @@
...
+  | 'string';
 
 /**
  * Represents an interface for adding and retrieving values to open graph objects and actions.
  */
 class ShareOpenGraphValueContainer {
   _properties: OpenGraphProperties;
 
-  constructor(properties: ?OpenGraphProperties) {
+  constructor(properties: OpenGraphProperties | null) {
```
rnike, line 42: this should be `properties: OpenGraphProperties` according to previous type

src/FBProfile.ts
```diff
@@ -1,22 +1,19 @@
...
-type ProfileMap = {
-  firstName: string,
-  lastName: string,
-  middleName: string,
-  imageURL: string,
-  linkURL: string,
-  userID: string,
-  email: string,
-  name: string,
+export type ProfileMap = {
```
rnike, line 8: After looked into the source code of both `facebook-ios-sdk` and `facebook-android-sdk`, I think the correct type of each key is  `string | undefined | null`

rnike: fixed

rnike (commented):
self reviewed all files, issues found are all solved

mikehardy:
> It will be wonderful if there's a beta version on npm to test it properly, currently I am just pushing it to my repo and yarn add from there.
> 
> None is it necessary, just consider it as a `nice to have`

A strict requirement of the migration for me is that it is buildable from source, from the installed version or from a version that is in git (so people may use a repo or branch or commit ref) then they should be able to change into the node_modules/react-native-fbsdk-next directory and run `yarn` (or `yarn && yarn build` or whatever) and it works. If that's okay then it's okay to alpha test at least, you can put out a call for testers on the roadmap issue perhaps? I can test it as well. I'm still behind deadline on one project and now it's the holidays so I haven't had time personally but this is critical for the repository's future I think, so it *will* get reviewed and merged soon I hope :-), thanks for your patience

rnike:
I've created a pre-release named `v7.0.0-alpha.1` including the recent feature I've made in my repo [rnike/react-native-fbsdk#v7.0.0-alpha.1](https://github.com/rnike/react-native-fbsdk-next/releases/tag/v7.0.0-alpha.1), but what I really want to do is to publish `v7.0.0-alpha.1` to npm for a proper testing, @mikehardy any idea how can I achieve that?

mikehardy:
I can try after I review it. The holidays mean I'm quite short on time though, apologies

rnike:
@mikehardy That's ok, this is no hurry. Thank you for still replying during your holidays.

mikehardy:
status for me: 
1) I've pulled the branch and run the example app successfully, as a basic criteria for merge - everything is :100: 
2) I want to review the actual changes but may just not have time - so the strategy I will actually use is exactly as @rnike suggests - I'll see about getting a beta version out, then people may test integrate it. We can put a time box around how long we wait prior to full release perhaps? Maybe at the end of the first week of January? In the end people do not *have* to upgrade and we'll mark it as a major version just in case so it should be okay either way

mikehardy (approved):
Merging this into a new beta branch, and semantic-release should auto-magically publish that in to a beta tag on npmjs.com

mikehardy:
I'm having a really frustrating time getting the version and tag lined up the way I like.
I have so far been unable to trigger a forced breaking change, and I think I know why but I am out of time for now on holidays to fix it.

The release is officially out there on the `beta` tag on npmjs.com so anyone may install it. It is still in the v6 series though, which is unwanted. It is released from the beta branch here until merged back to master. beta branch is strictly ahead of master at the moment so that part is easy to manage.

Future steps:
- add configuration to semantic release on beta branch so it understands the `!` syntax in commit messages for breaking changes (I did this in some other repo, I think react-native-google-ads so can copy from there)
- make another `!` commit on beta branch and trigger release action on beta branch so beta tag on npmjs.com lines up with a v7 beta release
- test integrate beta package with various work projects and my react-native-authentication-example-template to make sure it's good
- let it sit until january then merge to master and release there ?

mikehardy:
This is now successfully out as v7.0.0-beta.1 on npmjs.com and on the beta branch, got it all working.
Now it can incubate - and this is a call for testers, any interested parties should test this in their projects and report problems before end of first week of January!

rnike:
Thanks

Issue #159, opened 2021-12-22
Update deprecated native method


rnike:
Some of the methods are deprecated and will be removed in the next major release (or perhaps removed already)

**iOS issue**
- [x] ~change FBSDKAppEvents to FBSDKAppEvents.shared (#160)~

**Android issues**
- [x] ~AppEventsConstants.EVENT_NAME_PURCHASED is deprecated, [the doc](https://javadoc.io/static/com.facebook.android/facebook-core/5.9.0/com/facebook/appevents/AppEventsConstants.html#EVENT_NAME_PURCHASED) told us to use Use `AppEventsLogger.logPurchase` instead.~ https://github.com/thebergamo/react-native-fbsdk-next/blob/331fc983d21383230e6f703b01f5d1b53b8fd178/android/src/main/java/com/facebook/reactnative/androidsdk/FBAppEventsLoggerModule.java#L373 **update:** the reason why it's deprecated is because `logPurchase` is a shortcut for logging this event(metioned [here](https://github.com/facebook/facebook-android-sdk/blob/5fe6e2a9d7056a17f54c1cae13e00788723d34f6/facebook-core/src/main/java/com/facebook/appevents/AppEventsConstants.kt#L78-L83)), according to that, this event will probably not be removed and should be a won't fix.
- [x] ~fix several `access static method via instance` issues in `FBAppEventsLoggerModule.java`~ (not critical, won't fix)
- [x] ~`GameRequestDialog` in `FBGameRequestDialogModule.java` is deprecated use `com.facebook.gamingservices.GameRequestDialog` instead~ (#172)
- [ ] `MessageDialog` in `FBMessageDialogModule.java` is deprecated
- [ ] `SendButton` in `FBSendButtonManager.java` is deprecated. **(update 1/4)** According to https://github.com/facebook/facebook-android-sdk/commit/ab4f0c8867f40acf204d208af884836c38b935fb, Sharing to Messenger via the SDK is unsupported **[need investigation]**
- [ ] `promise.reject(String message)` is deprecated in `FBShareDialogModule.java`
- [x] ~`RCTEventEmitter` in `RCTLoginButton.java` is deprecated~ (won't fix according to https://github.com/thebergamo/react-native-fbsdk-next/pull/163#discussion_r776561471)
- [x] static method `buildAppInviteContent` in `Utility.java` has deprecated calls (removed in #176)
- [ ] static method `buildShareLinkContent` in `Utility.java` has deprecated calls

PR #160, opened 2021-12-22
fix(iOS)!: update deprecated methods

rnike:
According to #159, there're some deprecated methods still in use,

luckily on iOS, we'll only have to change `FBSDKAppEvents` to `FBSDKAppEvents.shared`,

still having trouble developing android, which will take me a while, so this pr is only going to fix iOS


mikehardy (approved):
merging this in to beta branch as well, the previous commit on that branch did not trigger release perhaps because it was "refactor"

PR #165, opened 2021-12-29
feat(ios): AEM logger

### LINKED ISSUE #153: AEM logging implementation for iOS
Hi, currently on my work we need the AEM Logging feature for iOS but it isn't in this sdk yet.

Offical document: [Aggregated Event Measurement(AEM)](https://developers.facebook.com/docs/app-events/guides/aggregated-event-measurement/)

Here's what I'm planning to do

1. export the `FBAEMReporter recordAndUpdateEvent:currency:value:parameters:` function to js (mentioned in the document at **Integration with Aggregated Events Measurement Kit -> Step 4. Add AEM Logging**)
2. documentation for AEM

rnike:
close #153

[Aggregated Event Measurement for Facebook App Events](https://developers.facebook.com/docs/app-events/guides/aggregated-event-measurement/)

1. fix project file issue in `project.pbxproj` (not a serious issue, it's just cannot open some files with xcode when developing this project)
2. create `FBAEMReporter.logAEMEvent` method
3. add a section for AEM in README.md
a. The official document told us to `Add the following to your Podfile: pod 'FBAEMKit'`, but in our case, `FBAEMKit` is already a dependency for `FBSDKCoreKit`, so I didn't mention it in README.md

### DIFF
--- a/src/FBAEMReporter.ts
+++ b/src/FBAEMReporter.ts
@@ -1,0 +1,45 @@
+/**
+ * This source code is licensed under the license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import {NativeModules, Platform} from 'react-native';
+
+/**
+ * Aggregated Event Measurement (AEM) for iOS apps allows for the measurement of app events
+ * from iOS 14.5+ users who have opted out of app tracking.
+ *
+ * @ref https://developers.facebook.com/docs/app-events/guides/aggregated-event-measurement/
+ */
+export default {
+  /**
+   * Log AEM event, event names for AEM must match event names you used in app event logging.
+   *
+   * *This method will bypass if platform isn't iOS*
+   *
+   * **Make sure you also handle the DeepLink URL with AEM**
+   *
+   * @ref https://developers.facebook.com/docs/app-events/guides/aggregated-event-measurement/
+   * @platform iOS
+   */
+  logAEMEvent: (
+    eventName: string,
+    value: number,
+    currency?: string,
+    otherParameters?: Record<string, string | number>,
+  ) => {
+    // AEM is currently only for iOS
+    if (Platform.OS !== 'ios') {
+      return;
+    }
+
+    NativeModules.FBSDKAEMReporter.recordAndUpdateEvent(
+      eventName,
+      value,
+      currency,
+      otherParameters,
+    );
+  },
+};
--- a/README.md
+++ b/README.md
@@ -633,13 +633,13 @@ The DeepLink URL from the re-engagement ads should be passed to the AEM Kit even
 
 #### **Step 2. Add AEM Logging**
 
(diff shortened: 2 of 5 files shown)


src/FBAEMReporter.ts
```diff
@@ -0,0 +1,45 @@
...
+   * Log AEM event, event names for AEM must match event names you used in app event logging.
+   *
+   * *This method will bypass if platform isn't iOS*
+   *
+   * **Make sure you also handle the DeepLink URL with AEM**
+   *
+   * @ref https://developers.facebook.com/docs/app-events/guides/aggregated-event-measurement/
+   * @platform iOS
+   */
+  logAEMEvent: (
```
thebergamo, line 27: Reviewing from mobile, but isn't it only iOS available?
In case it won't be ported to android in near future, I would make it platform specific as we did for other APIs instead of fallback in android, just for sake of consistency.

In case we do have an android version of it and it's just a case of porting it I would make it clear that support for Android would come soon and/or we're open for contributors :)

thebergamo (commented):
Besides my comment it LGTM

src/FBAEMReporter.ts
rnike: Yes, it's iOS only feature. I'm not sure what you mean by "make it platform specific instead of fallback in android", I thought marking it as `@platform iOS` is what we do for platform specific, the actual use case, as far as I know, is to send an additional event using logAEMEvent

```js
  AppEventsLogger.logEvent(eventName, value, parameters);
  AEMReporter.logAEMEvent(eventName, value, currency, parameters);
```
The early return on the none-ios platform is to make the function usage easier, so we don't have to do something as below
```js
  AppEventsLogger.logEvent(eventName, value, parameters);
  if(Platform.OS === 'ios'){
    AEMReporter.logAEMEvent(eventName, value, currency, parameters);
  }
```
I don't know if there will be an android version of AEM, not mentioned on the document, but I think it will not be happening in the near future.

thebergamo: I got your point, problem is that someone implementing it could be confused and expect that it would works for android as well.

https://developers.facebook.com/docs/facebook-login/limited-login/ios

Take a look in the above part in our Readme.md

thebergamo: Sorry I meant this method: loginTrackingIOS

NB!: the iOS app doesn't help in copy/paste some internal links :(

rnike: Taking your advice, I renamed it to `AEMReporterIOS.logAEMEvent`

mikehardy (approved):
This looks good to me @thebergamo if this looks good to you, then if you merge it will go into the beta branch (I set the target to beta), and then running the release action *on the beta branch target* will generate another beta I believe

Once both have "marinated" (this one, and typescript) we can then do a PR from beta to master and do a real v7 release ?

thebergamo:
Seems great now, merged o/

PR #168, opened 2022-01-04
fix(example)!: Copy package.json to example on example:devcopy

rnike:
Currently working on the example project in this repo is a nightmare, I'm trying to change the example into TypeScript but facing some problems.

One of the problems is the file path in the `package.json` is wrong, adding `package.json` to the `example:devcopy` should fix it.


rnike:
Closing it due to it's included in #169

PR #169, opened 2022-01-04
feat!: Example rework

rnike:
Example rework, changing to Typescript. 

Previously it was using flow which is not existing anymore in this repo.

This will make it easier to contribute and write example code.

The further things we can do in the example project, and of course only if needed
1. Add more examples
2. Unit test improvement


package.json
```diff
@@ -115,25 +115,19 @@
...
-    "source": "src",
-    "output": "lib",
-    "targets": [
-      [
-        "commonjs"
-      ],
-      "module",
-      "typescript"
-    ]
-  },
```
rnike, line 128: Removing `@react-native-community/bob` because we are not using it

src/index.ts
```diff
@@ -39,3 +39,8 @@ export {default as LoginButton} from './FBLoginButton';
 export {default as SendButton} from './FBSendButton';
 export {default as ShareButton} from './FBShareButton';
 export {default as AEMReporterIOS} from './FBAEMReporter';
+
+export type {
+  ShareContent,
+  ShareContentCommonParameters,
+} from './models/FBShareContent';
```
rnike, line 46: Currently, we need the type `ShareContent` in the example project to avoid type error.

We might be missing more types to export excepted `ShareContent` and `ShareContentCommonParameters`, my strategy here is to only fix issues that I have encountered.

tsconfig.build.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "./tsconfig",
+    "exclude": [
+        "RNFBSDKExample"
+    ]
```
rnike, line 5: Add `"RNFBSDKExample"` to `exclude` for typescript build otherwise it will build files end with `.ts`/.`tsx` in the example which is not expected.

package.json
```diff
@@ -60,7 +60,7 @@
     "example:android": "cd ./RNFBSDKExample && yarn android --no-jetifier",
     "example:clean": "cd RNFBSDKExample && \\rm -fr yarn.lock node_modules ios/Podfile.lock && cd ..",
     "example:install": "cd RNFBSDKExample && yarn && cd ios && (rm -f Podfile.lock && pod install || true) && cd ../..",
-    "example:devcopy": "yarn prepare && cp -rv android ios lib src *.podspec RNFBSDKExample/node_modules/react-native-fbsdk-next/",
+    "example:devcopy": "yarn prepare && cp -rv android ios lib src package.json *.podspec RNFBSDKExample/node_modules/react-native-fbsdk-next/",
```
rnike, line 63: As mentioned in #168, this will fix the wrong file path pointing from `node_modules/react-native-fbsdk-next/package.json`

package.json
```diff
@@ -115,25 +115,19 @@
     "bracketSameLine": true,
     "parser": "typescript"
   },
-  "@react-native-community/bob": {
```
rnike, line 118: Removing @react-native-community/bob because we are not using it

mikehardy:
Not sure how others feel about this but examples to me are throw-away. Almost all work on them - and certainly infrastructure work - is wasted time. Anything you do here will require maintenance going forward and you will be limiting your ability to provide real value by reducing time available since you are working on examples

So I strongly don't want this. I want this https://github.com/thebergamo/react-native-fbsdk-next/blob/beta/refresh-example.sh to be maintained instead, personally.

it serves as a demonstration of integration (frequently required when trying to get someone that logged an issue to actually demonstrate clean reproduction of the issue) and it also means there is no "upgrade" cost as react-native inevitably issues new releases, like 0.67 which will release shortly and require more work on these same files you are touching here

If touching anything inside the example directory can be avoided, it should be avoided I think.

If you really really want typescript in the example, I propose modifying the script to build off the community typescript template as a start, however that will actually *harm* the usefulness of the example as javascript people and typescript people don't overlap, but typescript people (myself included) are used to handling javascript - but the inverse is not true

rnike:
@mikehardy I never think about what the example actually is before, thought it was a kind of place for demonstration or testing, but I got your concept, it's just for CI app build(correct me if I'm wrong).

I'll take a look at `refresh-example.sh`, closing this as well.

mikehardy:
It is definitely for demonstration as one of it's purposes, and the refresh script *does* take the existing App.js and other files, save them on the side, then rebuild the whole scaffolding before dropping them back in. So some state is preserved across refreshes, and that state is used to demonstrate functionality

But! It's more than that yes, it is also for CI test compile purposes, the refresh script itself demonstrates how to integrate the module, step by step, and may be used to create a clean repro for people, and there are costs associated with the code simply existing, the primary cost being that keeping it up to date vs react-native updates can be time consuming

For all those reasons, I usually look for fully automated example app construction (reduce maintenance cost, demonstrate integration) and I usually look for the widest possible demonstration of functionality based on the most widely recognized template (so, javascript, based off the react-native default template)

In practice with this strategy in use for a couple years in various repos it has been really really useful - the example is still useful as an example of functionality but maintenance of the app and triage of issues is lower cost in practice

rnike:
Got it. At the first, I saw some issues in the example(e.g. file path in package.json), and App.js is kind of outdated, but I'm definitely doing it overdesigned by driving it to REWORK, also changing it to TypeScript is not a good idea.

I'll manage to pick out the issues I saw previously and fix them individually, this will make things more reasonable, thank you @mikehardy for pointing this out.

PR #172, opened 2022-01-07
fix: Replace deprecated GameRequestDialog class

rnike:
As mentioned in [the facebook-android-sdk source code](https://github.com/facebook/facebook-android-sdk/blob/main/facebook-share/src/main/java/com/facebook/share/widget/GameRequestDialog.java#L50) ，`com.facebook.share.widget.GameRequestDialog` is deprecated should use `com.facebook.gamingservices.GameRequestDialog` instead

**Test Plan**

`GameRequestDialog.canShow` and `GameRequestDialog.show` works fine on android app.

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBGameRequestDialogModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBGameRequestDialogModule.java
@@ -28,7 +28,7 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.module.annotations.ReactModule;
 import com.facebook.share.model.GameRequestContent;
-import com.facebook.share.widget.GameRequestDialog;
+import com.facebook.gamingservices.GameRequestDialog;
 
 /**
  * Provides functionality to send requests in games.


mikehardy (commented):
Nice - LGTM but this mixes a deprecation fix (changing the object in use) which should be a non-functional / non-semantic change with what appears to be a semantic change (altering how the error is reported when promise is rejected).

That makes it more difficult to review, where deprecation is typically just "approve + merge + release"

I checked the promise implementation to see what the functional change means - https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/bridge/Promise.java

So it appears you have switched a return value from a "generic code" to a specific code now. If you think cautiously, that may actually be a breaking change - when APIs change their return values you may end up having the return value no longer match consuming library's code-matching conditional structures, right?

I'm not sure it's worth the potential risk to mix the two things. Possibly better to log a cleanup issue for next major/breaking release that changes *all* promise rejections to have named codes, if that's a priority, but delay that part for now just in case. "Possibly better" in this case is judged by whether this module gets a reputation for quietly releasing breaking changes in patch releases that don't affect most people but do bite a couple or not...

rnike:
Sure, I shouldn't mix up with `the promise.reject` in this PR. The meaning of that change is not what I thought that simple, removed from this PR, will look back into this afterward.

mikehardy (approved):
Quick and easy :smile: :sunglasses:

PR #173, opened 2022-01-07
feat: Export all types

rnike:
Don't want anyone having type trouble using this package, simply export all types we have declared as export from all files.


mikehardy (approved):
Ah yes - I totally agree with you here. It's always frustrating when I have to reach deeply into a package to get some type, I'm never sure why some packages don't export all the things? Good thinking

PR #175, opened 2022-01-07
test(example): quiet example build, update deps, use beta for example

mikehardy:
Each commit has its own notes which are useful to read, I encourage commit by commit review

The refresh-example commits are just the result of running the script, they should be uninteresting - the changes made in the commit just before re-running the script contains the actual changes

All of this was done with the goal of making sure the example still worked with the beta branch before we release it, and I just did some janitor work while I was in there

The beta branch works though! :-). Going to test it in one of my apps, and if it works I think we are good to release v4


rnike (approved):
Looks great to me, learned a lot from your commits, thanks. 

Besides, I checkout on this branch and run `refresh-example.sh` turns out diffs in `./RNFBSDKExample/ios/RNFBSDKExample.xcodeproj/project.pbxproj`, which is kind of weird, but probably it's because of the different Xcode versions, not an issue I think after `yarn example:ios` run successfully on my mac.

mikehardy:
Yeah, the pbxproj file is a little unstable. Sometimes it will change simply by re-running pod install. It's one of the reasons I just don't like Xcode as a programming environment - the way things are configured is only barely human-readable, and was never meant to be altered programmatically. But it is something we have to deal with :shrug:

mikehardy:
I should mention - that instability on the Xcode files is one of the main reasons I make the script changes in one commit with an explanation, then one completely separate commit that is nothing but the script run results. So that there is no confusion on which thing did what - the script run commits should have no manual changes in there if I'm doing it right

PR #176, opened 2022-01-10
fix(android): remove dead code

rnike:
Willing to fix deprecated calls in `buildAppInviteContent` mentioned in #159, but found out that `buildAppInviteContent` is no longer in use. 

Also found `buildAppGroupCreationContent` is not in use.

Remove `buildAppInviteContent` and `buildAppGroupCreationContent` from Utility.java in this PR.

**Test Plain**
Successfully build an example app for Android.


mikehardy (approved):
It wouldn't build if it didn't work and there's no chance these were reflectively used because the JS/Java bridge doesn't work that way with static methods. Thanks @rnike ! 

Just slightly related, I want to do one more quick check with a work project here on the beta then I'm hoping for release today or tomorrow, following our previously discussed timeline. I want to get this new version out there. Cheers!

PR #179, opened 2022-01-12
Fix in IOS Facebook Login when Facebook App Installed

### LINKED ISSUE #59: IOS: The FBSDK not work when facebook app is installed
check this issue https://github.com/facebookarchive/react-native-fbsdk/issues/785 

FBSDK not work when facebook app is installed

devmindset-shaunvos:
Fix in IOS Facebook Login when Facebook App Installed:
https://github.com/thebergamo/react-native-fbsdk-next/issues/59#issuecomment-1010838962

I experienced the same issue in IOS 15.2 whereby my  new app would not login via Facebook if the Facebook app was already installed (it logged in when the Facebook app was not installed).

Explanation of fix:

If you open the file [src/FBLoginManager.js](https://github.com/thebergamo/react-native-fbsdk-next/blob/master/src/FBLoginManager.js) you will see the enumerated types of **LoginBehaviorAndroid** and **LoginBehaviorIOS**. 

```
export type LoginBehaviorAndroid =
  // Attempt login in using the Facebook App, and if that does not work fall back to web dialog auth.
  | 'native_with_fallback'
  // Only attempt to login using the Facebook App.
  | 'native_only'
  // Only the web dialog auth should be used.
  | 'web_only';
/**
 * Indicate how Facebook Login should be attempted on iOS.
 */
export type LoginBehaviorIOS =
  // Attempts log in through the Safari browser.
  // This is the only behavior supported by the native sdk.
  'browser';
```
For those who have installed the package in your React Native project you can find the same file under the directory: node_modules/react-native-fbsdk-next/src/FBLoginManager.js.

If you look at the [Facebook Android Documentation](https://developers.facebook.com/docs/reference/android/current/class/LoginBehavior) the enumerated type **LoginBehaviorAndroid** looks like it matches as required.

If you look at the [Facebook IOS Documentation](https://developers.facebook.com/docs/reference/ios/current/class/FBSDKLoginManager) the enumerated type **LoginBehaviorIOS** looks like it doesn't match as required.[](url)

Fixes #59  to automatically close the issue when the PR is merged.

Test Plan (apologies its not thorough - no time):

- Tested fix on React Native 0.67RC6 on IOS 15.2 Physical Device (iPad 8th generation)
- Tried with Facebook App installed and not installed.  Always defaults to using browser for login to Facebook (issue no longer occurs).

Personal motive:
I am new to React Native development (just under a year experience).  I have zero experience in Swift or Objective C so sorry if I have mistakenly followed wrong procedure.  I was just experiencing this bug for my app as well and decided to take on the challenge of resolving myself.  Many thanks to the package maintainers for their contributions.

### DIFF
--- a/src/FBLoginManager.js
+++ b/src/FBLoginManager.js
@@ -51,7 +51,8 @@ export type LoginBehaviorAndroid =
 export type LoginBehaviorIOS =
   // Attempts log in through the Safari browser.
   // This is the only behavior supported by the native sdk.
-  'browser';
+  | 'FBSDKLoginBehaviorNative'
+  | 'FBSDKLoginBehaviorWeb';
 /**
  * Shows the results of a login operation.
  */


rnike:
Hi, thanks for the work, but I don't think this will fix the issue, if you look into the code you can see the `getLoginBehavior` will always resolve `'browser'` on iOS,

https://github.com/thebergamo/react-native-fbsdk-next/blob/ba45840a06c1bd1fdb885d88ab05d8bd619aa629/src/FBLoginManager.js#L87-L96

and the `setLoginBehavior` is doing nothing on iOS too,
https://github.com/thebergamo/react-native-fbsdk-next/blob/ba45840a06c1bd1fdb885d88ab05d8bd619aa629/src/FBLoginManager.js#L101-L106

so I expect simply changing the type `LoginBehaviorIOS` won't fix anything and it might turn out some `type errors` (e.g. `browser ` is not a type of `LoginBehavior`).

Besides, I am not facing this login issue (or perhaps I didn't notice the issue exists) in our project, so currently I'm not putting effort into it, what I mention above if has any mistake, feel free to correct me.

rnike:
By the way, I encourage you to work on the branch based on the `beta` branch, because we are going to merge it into `mater` these days, everything might will have conflicts if it's based on the current `master` after the `beta` has merged.

devmindset-shaunvos:
> Hi, thanks for the work, but I don't think this will fix the issue, if you look into the code you can see the `getLoginBehavior` will always resolve `'browser'` on iOS,

All i can confirm is this change fixed my "issue".  I was rebuilding my app with a fresh install of React native 0.67RC6, I had installed the latest packages (following latest docs as best I could) and then moved my older JavaScript code over.  I have since been updating my older javascript code to use the latest packages. I use a Mac Mini M1 [Apple Silicon] (2020).

`getLoginBehavior()` looks to me  like it's may work for Android though I'm not sure about IOS? I can't find where its called. or could the export type be different to whats exported from module.exports?

if I go further into my Pods folder FBSDKLoginKit/XCFrameworks/FBSDKLogin.xcframework/ios-arm64_armv7/FBSDKLoginKit.framework/Headers/FBSDKLoginManage.h I only see it defined as follows:
```
@property (nonatomic, assign) LoginBehavior loginBehavior;
@property (nonatomic, assign) DefaultAudience defaultAudience;
```

export type LoginBehavior = LoginBehaviorIOS | LoginBehaviorAndroid;

/**
 * Indicate how Facebook Login should be attempted on Android.
 */
export type LoginBehaviorAndroid =
  // Attempt login in using the Facebook App, and if that does not work fall back to web dialog auth.
  | 'native_with_fallback'
  // Only attempt to login using the Facebook App.
  | 'native_only'
  // Only the web dialog auth should be used.
  | 'web_only';
/**
 * Indicate how Facebook Login should be attempted on iOS.
 */
export type LoginBehaviorIOS =
  // Attempts log in through the Safari browser.
  // This is the only behavior supported by the native sdk.
  //'browser';
  | 'FBSDKLoginBehaviorNative'
  | 'FBSDKLoginBehaviorWeb';
```

**Comment by mikehardy** [comment]
With sincere apologies, I've just merged the typescript rewrite and this was the only pull request outstanding but it definitely blew this one up, because it was changing the old javascript file. It seems like there was a legitimate question about whether this would work or not, but either way it will need an update to go against the new typescript file

as a separate matter - if you're using patch-package (something I do a lot) and it's against a typescript package (like this one) and you patch the typescript make sure to either run yarn + the build now inside the node_modules/react-native-fbsdk-next after the patch or take the patch after modifying typescript and running the build

If that doesn't work (the build, from inside node_modules after running yarn in there) please let me know, it's my specific intention that we are shipping all the files needed for people to hack around i node_modules/react-native-fbsdk-next so they can help themselves if needed, though there will be the transpile step now

**Comment by Gp2mv3** [comment]
Hello,
I think I found the solution. The issue on our side was corrected by adding a sceneDelegate to our app. Since I added the following code (found in the Installation instruction in Facebook's doc) in a sceneDelegate.swift file in our project, the login now works perfectly:
```swift
import FBSDKCoreKit

func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
    guard let url = URLContexts.first?.url else {
        return
    }

    ApplicationDelegate.shared.application(
        UIApplication.shared,
        open: url,
        sourceApplication: nil,
        annotation: [UIApplication.OpenURLOptionsKey.annotation]
    )
}
```

I hope it can help you with your issues. I also had the fix of this PR applied so the solution may be to have both.

harshilmobmaxime:
Screenshot 2022-12-26 at 12 10 51 PM
add this code at the end of AppDelegate.m file. it's fix for me

PR #180, opened 2022-01-12
feat: Merge 7.0.0-beta to master

rnike:
Hello, I think it's time to merge 7.0.0-beta to master and release 7.0.0, 

although it's a bit late it should be merged last week as scheduled, 

anyway, we have run the beta in production for a while (we are on 7.0.0-beta2) and it seems fine,

if any issue occurs after merged I'll be here to help.


mikehardy:
I agree with you - I am the delay - the only hold up is that when I test integrated it into my work project and one of the templates it appeared I had breaking changes with regard to where things are located. I'm just not sure if that's intended or not. Since it did not succeed immediately (I thought it would! I was surprised actually) I was not able to just say "go for it". If we have breaking changes on where things are located we must provide a migration of where each symbol moved. If we do not have breaking changes and my project is broken (a strong possibility) then I need to see it run.

Anyway, that's the delay, I'm sorry. I am actually working on almost nothing but this....except also my country's school system went back to virtual classes and now I'm also the science and math tutor at home which is just killing my productivity. All I can do is ask for patience, sorry

rnike:
@mikehardy No need to be sorry you've done a lot, I appreciated it. Actually, my first thought is me causing the delay, I was wondering if there's anything I've missed to do, so that's why I'm opening this pr. I'll change this to draft and wait for now, or if you want it to be closed, is okay to let me know. Family, body care and work are more important than an open-source project, so just take your time, I'll be patient with it, also feel free to let me know if anything I can do for help.

mikehardy:
Okay - so I got a chance to try this on an old project - I just swapped the existing v6 fbsdk-next for a dependency on `beta`, I get this:

```
src/screens/PopupTest.tsx(10,31): error TS2307: Cannot find module 'react-native-fbsdk-next/types/FBAccessToken' or its corresponding type declarations.
src/stores/UserStore.ts(1312,9): error TS2345: Argument of type 'null' is not assignable to parameter of type 'GraphRequestConfig | undefined'.
```

So some of the symbols have moved, but we are not providing any guidance to developers on what symbols moved and where they moved to.

We either need to provide a map for developers of old import locations to new ones, or we need to provide the symbols in the old locations so it's seemless

@rnike do you know which symbols moved? Is it possible to make the symbol locations backwards-compatible?

mikehardy:
Also got this one - perhaps the typing is more correct now though?

```
Argument of type '(error: any, result: object | PromiseLike<object>) => void' is not assignable to parameter of type 'GraphRequestCallback'.
  Types of parameters 'result' and 'result' are incompatible.
    Type 'Record<string, unknown> | undefined' is not assignable to type 'object | PromiseLike<object>'.
      Type 'undefined' is not assignable to type 'object | PromiseLike<object>'.
```

mikehardy:
I think both of the things I saw were objectively my fault, they were me trying some random typings in a project-specific way because the typings weren't great before, but now they are much better. I think it's time

mikehardy (approved):
Integrated with a work project and it all works after I fixed a couple bogus project-specific typings

mikehardy:
It's out! @rnike this is so cool, thank you again for doing it, and thank you so much for the patience for review and release and stuff. I really hate being a barrier on progress and I'm glad this is out of the way. Going forward if you see pull requests that look obviously good, and you're doing your best, hit that publish button. I don't want to be a barrier ever really, and it's only the really big stuff (like this one) that I think need lots of collaboration. That's my philosophy anyway, in the repos I work on. Cheers!

PR #184, opened 2022-01-16
fix: delete jcenter use to solve problem with outage and future sunset

osvald0:
Jcenter is down now ([Check status here](https://status.bintray.com)).
It isn't related to the [library sunset](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) because it will be in read mode for a long time but we should replace it.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,8 +8,9 @@ buildscript {
     // module dependency in an application project.
     if (project == rootProject) {
         repositories {
+            mavenCentral()
             google()
-            jcenter()
+            gradlePluginPortal()
         }
         dependencies {
             classpath 'com.android.tools.build:gradle:7.0.3'
@@ -42,6 +43,7 @@ repositories {
     }
     mavenCentral()
     google()
+    gradlePluginPortal()
 }
 
 def FACEBOOK_SDK_VERSION = safeExtGet('facebookSdkVersion', '12.+')
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,7 +10,6 @@ buildscript {
         repositories {
             mavenCentral()
             google()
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
interesting! note that from the point of view of this module being used in other projects, the `(project == rootProject)` test will return false, so that jcenter entry will never be seen by apps integrating this, so it's good to remove it but won't actually fix people. It's a good change, just not a fix for build errors.

The gradlePluginPortal...I'm not so sure on that one - I don't think you want to add that. I think it will actually make things worse. https://github.com/react-native-community/discussions-and-proposals/issues/331#issuecomment-1013339743

osvald0:
Yes, make sense based on that comment. You can delete this PR then.

Thanks for the clarification!

android/build.gradle
```diff
@@ -42,6 +43,7 @@ repositories {
     }
     mavenCentral()
     google()
+    gradlePluginPortal()
```
mikehardy, line 46:
```suggestion
```

android/build.gradle
```diff
@@ -8,8 +8,9 @@ buildscript {
     // module dependency in an application project.
     if (project == rootProject) {
         repositories {
+            mavenCentral()
             google()
-            jcenter()
+            gradlePluginPortal()
```
mikehardy, line 13:
```suggestion
```

mikehardy (approved):
I think there is still utility in purging jcenter from the build.gradle actually and I appreciate you posting this, I just removed the gradle plugin repo entries and then this seems like a good change to me

I'm going to retarget it towards the beta branch, as we are working towards a release there with a rewrite in typescript. Any day now I hope...

PR #191, opened 2022-01-22
fix(test): safely dereference NativeModule for test compat

### LINKED ISSUE #188: jest mocks insufficient after v7 typescript merge
```
  ● Test suite failed to run

    TypeError: Cannot read property 'getConstants' of undefined

      at Object.<anonymous> (../../../../node_modules/react-native-fbsdk-next/lib/commonjs/FBAppEventsLogger.ts:137:1)
      at Object.<anonymous> (../../../../node_modules/react-native-fbsdk-next/lib/commonjs/index.ts:36:1)
```

The FBAppEventsLogger.ts file grabs the native module (which does not exist in a jest context) and without checking it for null just dereferences into it to get constants, causing jest tests to fail

Either the native module needs to be mocked in the provided jest config, or the native module needs to be checked for undefined in the code, I think

mikehardy:
This allows the module to work unmocked in a jest context

Fixes #188

### DIFF
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -134,13 +134,13 @@ export type AppEventParam = {
   ValueYes: string;
 };
 
-const {
-  AppEvents,
-  AppEventParams,
-}: {
-  AppEvents: AppEvent;
-  AppEventParams: AppEventParam;
-} = AppEventsLogger.getConstants();
+let AppEvents: AppEvent | undefined = undefined;
+let AppEventParams: AppEventParam | undefined = undefined;
+if (AppEventsLogger !== undefined) {
+  const constants = AppEventsLogger?.getConstants();
+  AppEvents = constants.AppEvents;
+  AppEventParams = constants.AppEventParams;
+}
 
 export default {
   /**


src/FBAppEventsLogger.ts
```diff
@@ -134,13 +134,13 @@ export type AppEventParam = {
...
-  AppEvents: AppEvent;
-  AppEventParams: AppEventParam;
-} = AppEventsLogger.getConstants();
+let AppEvents: AppEvent | undefined = undefined;
+let AppEventParams: AppEventParam | undefined = undefined;
+if (AppEventsLogger !== undefined) {
+  const constants = AppEventsLogger?.getConstants();
+  AppEvents = constants.AppEvents;
+  AppEventParams = constants.AppEventParams;
+}
```
rnike, line 143: const { AppEvents, AppEventParams } = AppEventsLogger?.getConstants() || {};

mikehardy: @rnike if you want to make a PR that is cleaner (this looks cleaner...yes) please feel free, I had pretty limited time this weekend and it was breaking one of my pre-commit hooks in a work repo so I just patched as quickly as possible, maybe not as well as possible. Hopefully at least I just didn't break anything :crossed_fingers:

rnike: Yep don't worry the change you've made is not going to break anything, but I'm still wondering why jest is falling without this change, and why the `6.x.x` is not failing without it. I will make the refactor PR after I figure it out.

PR #193, opened 2022-01-24
refactor: make the `getConstants` call cleaner

rnike:
This makes the `AppEventsLogger.getConstants` code cleaner without changing functionality. 

--
Besides, I've made a few tests, the jest will still fail with the error message

 > Invariant Violation: `new NativeEventEmitter()` requires a non-null argument. 

that means the user of this package will need to add 

> jest.mock('react-native/Libraries/EventEmitter/NativeEventEmitter');

in their jest file.

But after that, if tend to call `AppEventsLogger.logEvent`  in jest will still fail at 

https://github.com/thebergamo/react-native-fbsdk-next/blob/a6c488ee85ebd91ca2b2f203f8d0cf49a937cb2b/src/FBAppEventsLogger.ts#L172

because AppEventsLogger is `undefined` after all.

The simplest way for now I think is to just mock the whole package

> jest.mock('react-native-fbsdk-next)

but in my opinion, for this package's completion, formally having a jest setup is still a `nice to have`, @mikehardy how do you think? I'm willing to work on it, still need some surveys before starting, 

1. Find out why adding `'<rootDir>/node_modules/react-native-fbsdk-next/jest/setup.js` is not working and try fixing it.
or
2. Use `react-native-gesture-handler` method -> https://github.com/software-mansion/react-native-gesture-handler/blob/main/jestSetup.js
or
3. Suggestions are welcome

### DIFF
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -134,13 +134,7 @@ export type AppEventParam = {
   ValueYes: string;
 };
 
-let AppEvents: AppEvent | undefined = undefined;
-let AppEventParams: AppEventParam | undefined = undefined;
-if (AppEventsLogger !== undefined) {
-  const constants = AppEventsLogger?.getConstants();
-  AppEvents = constants.AppEvents;
-  AppEventParams = constants.AppEventParams;
-}
+const {AppEvents, AppEventParams} = AppEventsLogger?.getConstants() || {};
 
 export default {
   /**


src/FBAppEventsLogger.ts
```diff
@@ -134,13 +134,7 @@ export type AppEventParam = {
...
 };
 
-let AppEvents: AppEvent | undefined = undefined;
-let AppEventParams: AppEventParam | undefined = undefined;
-if (AppEventsLogger !== undefined) {
-  const constants = AppEventsLogger?.getConstants();
-  AppEvents = constants.AppEvents;
-  AppEventParams = constants.AppEventParams;
-}
+const {AppEvents, AppEventParams} = AppEventsLogger?.getConstants() || {};
```
mikehardy, line 137: This is not a firm suggestion, just a typescript thought - would nullish coalescing be better here?

```suggestion
const {AppEvents, AppEventParams} = AppEventsLogger?.getConstants() ?? {};
```

rnike: It depends, in this circumstance  I'm using `||` for a reason, in the image below you can see `||` generates lesser js code than `??`, so usually in my work we use `||` instead of `??`. 

example

But `??` is still useful in some scenarios, like if the return value is falsely

```
// data[key] will probably be 0 but we only want defaultValue if it's undefined
data[key] ?? defaultValue
```

mikehardy: works for me :-), thanks for the explanation

mikehardy:
> Find out why adding '<rootDir>/node_modules/react-native-fbsdk-next/jest/setup.js is not working and try fixing it.

I think this is the best way, since that really should work shouldn't it? No idea why it is not :thinking:

PR #196, opened 2022-02-03
docs: include example of how to use AEM event logger

nicknjpconsultingllc:
I was a bit confused when I first saw the new AEM log function - it wasn't clear if I had to use it for purchases as well as regular events, and if so, what event name to use.

So I added some more explanation. Hopefully this is helpful.

One thing that I didn't know how to do was convert the logEvent args (Array<number | Params>) into the Record<string, string | number> type needed for logAEMEvent. Any help on that would be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -652,9 +652,9 @@ LogFBPurchase = (purchaseAmount: number, currencyCode: string, parameters?: Para
     AEMReporterIOS.logAEMEvent("fb_mobile_purchase", purchaseAmount, currencyCode, parameters);
 }
 
-LogFBEvent = (eventName: string, ...args: Array<number | Params>) => {
-    AppEventsLogger.logEvent(eventName, ...args);
-    AEMReporterIOS.logAEMEvent(eventName, 0, undefined, {});
+LogFBEvent = (eventName: string, valueToSum: number, parameters: Record<string,string | number>)=> {
+    AppEventsLogger.logEvent(eventName, valueToSum, parameters);
+    AEMReporterIOS.logAEMEvent(eventName, valueToSum, parameters.fb_currency, parameters);
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -644,6 +644,21 @@ AEMReporterIOS.logAEMEvent(eventName, value, currency, otherParameters);
 
 Event names for AEM must match event names you used in app event logging.
 
+Here's an example of how to use this method - 
+
+```ts
+LogFBPurchase = (purchaseAmount: number, currencyCode: string, parameters?: Params | undefined) => {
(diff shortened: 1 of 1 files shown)


nicknjpconsultingllc:
Tagging @rnike for confirmation that this really is the best way to use this new method.

README.md
```diff
@@ -644,6 +644,21 @@ AEMReporterIOS.logAEMEvent(eventName, value, currency, otherParameters);
...
+
+```ts
+LogFBPurchase = (purchaseAmount: number, currencyCode: string, parameters?: Params | undefined) => {
+    AppEventsLogger.logPurchase(purchaseAmount, currencyCode, parameters);
+    AEMReporterIOS.logAEMEvent("fb_mobile_purchase", purchaseAmount, currencyCode, parameters);
+}
+
+LogFBEvent = (eventName: string, ...args: Array<number | Params>) => {
+    AppEventsLogger.logEvent(eventName, ...args);
+    AEMReporterIOS.logAEMEvent(eventName, 0, undefined, {});
```
rnike, line 657: The AEM event will lose parameters on this line, in our project we implement it as below

```ts
export function LogFBEvent(eventName, data){
  // we have a create function to make payload from data by event name
  const payload = makePayloadFromDataByEvent(eventName, data);
  const { value, parameters } = payload;

  // the original log event we have already implemented before AEM exists
  AppEventsLogger.logEvent(eventName, value, parameters);

  // additionally send AEM event with payload same with previous log event
  AEMReporter.logAEMEvent(eventName, value, parameters.fb_currency, parameters);
}
```

if we just look at the types, it's hard to sync the usage because the interface between `logEvent` and `logAEMEvent` is quite different,  but if we look into the implementation we can see they are quite the same

I encourage using the interface below for `logEvent`

```ts
logEvent(eventName: string, valueToSum: number, parameters: {[key:string]:string|number});
```

because the only difference between them is `currency`

```ts
logAEMEvent(eventName: string, value: number, currency?: string | undefined, otherParameters?: Record<string, string | number> | undefined)
```

so after all, the `LogFBEvent` can be changed to

```ts
LogFBEvent = (eventName: string, valueToSum: number, parameters: Record<string,string | number>)=> {
    AppEventsLogger.logEvent(eventName, valueToSum, parameters);
    AEMReporterIOS.logAEMEvent(eventName, valueToSum, parameters.fb_currency, parameters);
}
```

hope I'm clear enough, any feedback is welcome

nicknjpconsultingllc: I like it! Changed!

rnike (approved):
LGTM

rnike:
Changed the title prefix to `docs:` from `fix:` according to the diff is about documentation improvement.

PR #204, opened 2022-02-21
feat: Add reauthorizeDataAccess method to LoginManager

neilco:
This pull request replicates the functionality added in facebookarchive/react-native-fbsdk#720 in order to facilitate the requesting of data access permissions when they have expired. The native Facebook SDK exposes a method on the LoginManager class (`reauthorizeDataAccess`) that the changes in this pull request exposes to the JS layer. 

Test Plan:

We've modified the example app to call the new method. We've also tested this with a separate app to verify the code is behaving as expected.

### DIFF
--- a/RNFBSDKExample/App.js
+++ b/RNFBSDKExample/App.js
@@ -23,7 +23,7 @@
 
 import React, {Component} from 'react';
 import {Alert, StyleSheet, Text, TouchableHighlight, View} from 'react-native';
-import {LoginButton, Settings, ShareDialog} from 'react-native-fbsdk-next';
+import {LoginButton, LoginManager, Settings, ShareDialog} from 'react-native-fbsdk-next';
 
 const SHARE_LINK_CONTENT = {
   contentType: 'link',
@@ -35,6 +35,15 @@ const SHARE_LINK_CONTENT = {
 Settings.initializeSDK();
 
 export default class App extends Component<{}> {
+  _reauthorizeDataAccess = async () => {
+    try {
+      const result = await LoginManager.reauthorizeDataAccess();
+      Alert.alert("Reauthorize data access result", JSON.stringify(result, null, 2));
+    } catch (error) {
+      Alert.alert("Reauthorize data access fail with error:", error);
+    }
+  };
+
   _shareLinkWithShareDialog = async () => {
     const canShow = await ShareDialog.canShow(SHARE_LINK_CONTENT);
     if (canShow) {
@@ -63,7 +72,10 @@ export default class App extends Component<{}> {
           }}
         />
         <TouchableHighlight onPress={this._shareLinkWithShareDialog}>
-          <Text style={styles.shareText}>Share link with ShareDialog</Text>
+          <Text style={styles.buttonText}>Share link with ShareDialog</Text>
+        </TouchableHighlight>
+        <TouchableHighlight onPress={this._reauthorizeDataAccess}>
+          <Text style={styles.buttonText}>Reauthorize Data Access</Text>
         </TouchableHighlight>
       </View>
     );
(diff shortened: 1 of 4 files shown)


RNFBSDKExample/App.js
```diff
@@ -77,7 +89,7 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     backgroundColor: '#F5FCFF',
   },
-  shareText: {
+  buttonText: {
```
mikehardy, line 92: nice touch, since it's two types of buttons now :+1:

ios/RCTFBSDK/login/RCTFBSDKLoginManager.m
```diff
@@ -90,14 +90,27 @@ + (BOOL)requiresMainQueueSetup
                                                     tracking: tracking
          ];
     }
-    
-    
+
+
```
mikehardy, line 94: not sure about the unrelated spacing changes, in most repositories I run I use the `clang-format` npm package to just enforce a consistent styles on the Obj-C code so no one has to think about it really (other than to turn the formatter) but not sure if I integrated it here, CI will tell the tale

mikehardy:
> [warn] Code style issues found in the above file(s). Forgot to run Prettier?

ios/RCTFBSDK/login/RCTFBSDKLoginManager.m
mikehardy: No problem with this one it appears I don't have clang-format integrated, and that's not on my priority list at the moment. Noted a typescript lint issue below though

ios/RCTFBSDK/login/RCTFBSDKLoginManager.m
```diff
@@ -90,14 +90,27 @@ + (BOOL)requiresMainQueueSetup
...
     [_loginManager
      logInFromViewController: nil
      configuration:configuration
      completion:requestHandler];
 };
 
+RCT_REMAP_METHOD(reauthorizeDataAccess, reauthorizeDataAccess_resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
```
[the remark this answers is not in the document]
neilco: `RCT_EXPORT_METHOD` uses a prefix mapping method to take everything from the ObjC signature up to the first colon. However, it's a common convention in ObjC to use a common prefix (e.g. the functions of AppDelegate that are prefixed `application:(UIApplication *)app`).

`RCT_REMAP_METHOD` allows you to rename the JS function to something unique when you are exporting an ObjC with a duplicate prefix. It's actually pretty handy. 

I've changed this in b3cf4ca to use `RCT_EXPORT_METHOD` as there are no prefix conflicts.

neilco:
> [warn] Code style issues found in the above file(s). Forgot to run Prettier?

You've got to love it when something tells you there's an issue, but no what the actual issue is. Maybe you need to tweak your prettier config? 

It's fixed in 8b8e9fc.

mikehardy:
> > [warn] Code style issues found in the above file(s). Forgot to run Prettier?
> 
> You've got to love it when something tells you there's an issue, but no what the actual issue is. Maybe you need to tweak your prettier config?
> 
> It's fixed in [8b8e9fc](https://github.com/thebergamo/react-native-fbsdk-next/commit/8b8e9fc46a959d4a3e39e2cb3ff42aeaab983824).

I think local prettier integrations (like the one in VScode at least, which you indicate you are using?) do look at the local .prettierrc and do tell you - either way because of the way prettier is, you just need to run it and that's that - but this is something that would be best in a pre-commit hook yeah - no reason to wait till it hits CI to see it

mikehardy (approved):
looks really good now, happy to approve + merge + release in just a moment, and though it doesn't matter now it appears that the alternative RCT_NNNNN method was just in case there was a symbol collision exporting the method? Interesting - makes sense I guess, just never seen it needed in any of the other repos I've worked on. Learn something every day...

PR #205, opened 2022-02-23
Expo plugin

### LINKED ISSUE #192: Support Expo EAS build via plugin config?
Hi, Would you like to support Expo EAS build via plugin config? I will send a PR for that support.

giautm:
Fixes #192

Test Plan:

- Unit-test on the plugin
- Integration test with an application that config the plugin

### DIFF
--- a/plugin/src/config.ts
+++ b/plugin/src/config.ts
@@ -1,0 +1,108 @@
+import { ExpoConfig } from '@expo/config-types';
+
+export type ExpoConfigFacebook = Pick<
+  ExpoConfig,
+  | 'facebookScheme'
+  | 'facebookAdvertiserIDCollectionEnabled'
+  | 'facebookAppId'
+  | 'facebookAutoInitEnabled'
+  | 'facebookAutoLogAppEventsEnabled'
+  | 'facebookDisplayName'
+>;
+
+export type ConfigProps = {
+  /**
+   * Used for all Facebook libraries. Set up your Facebook App ID at https://developers.facebook.com.
+   */
+  appID?: string;
+  /**
+   * Used for native Facebook login.
+   */
+  displayName?: string;
+  /**
+   * Used for Facebook native login. Starts with 'fb' and followed by a string of digits, like 'fb1234567890'. You can find your scheme [here](https://developers.facebook.com/docs/facebook-login/ios)in the 'Configuring Your info.plist' section (only applicable to standalone apps and custom Expo Go apps).
+   */
+  scheme?: string;
+
+  clientToken?: string;
+  /**
+   * Whether the Facebook SDK should be initialized automatically. The default in Expo (Client and in standalone apps) is `false`.
+   */
+  isAutoInitEnabled?: boolean;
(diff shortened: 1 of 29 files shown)


plugin/src/config.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Used for native Facebook login.
+   */
+  displayName?: string;
+  /**
+   * Used for Facebook native login. Starts with 'fb' and followed by a string of digits, like 'fb1234567890'. You can find your scheme [here](https://developers.facebook.com/docs/facebook-login/ios)in the 'Configuring Your info.plist' section (only applicable to standalone apps and custom Expo Go apps).
+   */
+  scheme?: string;
+
+  clientToken?: string;
```
mikehardy, line 27: I think it is important to at least mention that you pull this from the facebook developer dashboard as well, and that it will be required in the next major version of the facebook SDKs so you should configure it even if it is temporarily optional.

Thinking further: many of these ConfigProps are marked as optional, but, the appID and displayName and scheme etc - some of those are required and not optional, aren't they?

mikehardy (commented):
This looks great - I'll approve the CI run and hopefully it goes green.
I had a comment about the docs for the config elements and I'm curious for your thoughts there
It is really important to note that I do not use or test Expo functionality, so I will not be able to support this at all. I'm more than happy to have it in the library as it will open the library to a large pool of possibly interested users but I just won't be able to help any of those users with troubleshooting or maintain this code, so ongoing help would be more than appreciated. Is that something you can provide?

plugin/src/config.ts
giautm: I make some config are optional because I want to support Expo User moving from `expo-facebook` to this package. They ready has some config in the app.json so they don't have to declare it twice. But, I will add validation for missing `appID` and `displayName`. (`scheme` will default to `fb` + appID).

```
    facebookAppId,
    facebookDisplayName,
    facebookScheme,
    facebookAutoInitEnabled,
    facebookAutoLogAppEventsEnabled,
    facebookAdvertiserIDCollectionEnabled,
  ```

giautm:
> It is really important to note that I do not use or test Expo functionality, so I will not be able to support this at all. I'm more than happy to have it in the library as it will open the library to a large pool of possibly interested users but I just won't be able to help any of those users with troubleshooting or maintaining this code, so ongoing help would be more than appreciated. Is that something you can provide?

Yes, I can help to maintain this code/troubleshoot any issue with Expo.

mikehardy (approved):
Fantastic, and I can't tell you how much I appreciate the offer to help on maintaining it, I have limits to what I'm familiar with + have time to become familiar with and while I love the idea of Expo users being able to take advantage of the module I'm just at my limit now and don't personally don't know Expo :sweat_smile: 

This looks good to me now - I like the validation, let's see what CI says :-)

giautm:
I think the CI was green: https://github.com/thebergamo/react-native-fbsdk-next/actions/runs/1887573023

Issue #210, opened 2022-02-26
Build fails on iOS on 7.3.1 (M1 Silicon)


yemd:
I upgraded from 7.2.0 to 7.3.1, and it seems that the new Expo plugin is messing things up, 

## 🐛 Bug Report

I upgraded from 7.0.1 to 7.3.1, and it seems that the new Expo plugin is messing things up.

## Error
The following build commands failed:
        PhaseScriptExecution [CP-User]\ Generate\ Specs /Users/user/Library/Developer/Xcode/DerivedData/electra-efhopmcbavxoswdwpxqrefpzlblb/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/FBReactNativeSpec.build/Script-0C645E84D737C4B3EF4A9A92FB79B97C.sh (in target 'FBReactNativeSpec' from project 'Pods')
        CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler (in target 'ExpoAdapterFBSDKNext' from project 'Pods')
        CompileSwiftSources normal arm64 com.apple.xcode.tools.swift.compiler (in target 'ExpoAdapterFBSDKNext' from project 'Pods')

## To Reproduce
Create a new reactive native app, use expo modules, add fbsdk, build

## Expected Behavior
Build passes

## Environment
System:
    OS: macOS 12.1
    CPU: (8) arm64 Apple M1
    Memory: 100.47 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
    Watchman: 2022.02.21.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8193401
    Xcode: 13.2.1/13C100 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.12 - /opt/homebrew/opt/openjdk@11/bin/javac
  npmPackages:
[...]

PR #211, opened 2022-02-28
fix(expo, ios): remove unnecessary compiler_flags in plugin podspec

giautm:
Hi, I have a mistake in podspec. This commit will remove unnecessary compiler_flags


mikehardy (approved):
looks good to me - do you think this will have an effect on #210  ?

giautm:
> looks good to me - do you think this will have an effect on #210 ?

Nope, I don't think it effect #210 because I don't have enough information to find the root cause.

Issue #208, opened 2022-02-24
Update to v13 of native SDKs [testers wanted! patch-package patch ready!]


mikehardy:
facebook-android-sdk and facebook-ios-sdk just released v13 of their native SDKs

I'm happy maintaining this repo and do my best to merge PRs and fix little immediate issues but I do this in my free time

I do not have time to take on larger projects like forward porting to v13 right now

Anyone that needs v13 features or wants to see this library stay alive may consider either directly sponsoring me to get it done or should propose a PR themselves (which I will happily collaborate on!)

In the meanwhile, this issue will stay open for discussion among any interested parties, or status updates as they happen.

Please do not +1 or me too or ask for status though. Either contribute and move it forward, or quietly listen, to keep notification traffic down. It will be much appreciated. Thanks!

likon416:
Hi @mikehardy 

Yeah without SDK 11+ you can't really send events on iOS due to tracking permissions. So basically SDK9 is making this plugin (same as expo-facebook) obsolete. 
Can you point me to what needs to be done to update to a new SDK? Maybe I can help somehow.

mikehardy:
We're already on 12 here. Should be no issue. 13 was released last week, we should update over time but there should not be any functional problems with current versions used here

That said a PR to update would be great.

likon416:
ahh, great, that solves everything :) thanks

mikehardy:
Still doesn't solve that we need to move to v13 at some point ;-). But yes, you should not be blocked on anything. I'm going to hide these comments now as resolved to clean up the issue, no offense implied or intended, cheers !

kalyaniDM:
It is showing critical error to update facebook sdk from 12 to 13 in ios.Could you please help in updating the repository with updated facebook sdk version 13.1.0 .

mikehardy:
Could you please read my comment above and take appropriate action? I'll happily collaborate if you propose a PR.
This same comment applies to anyone else interested.

ashuOxit:
Hi @mikehardy 
i am able to update SDK but I am not able to update Gaming share functionality 

I am sharing a patch file you can check it and if possible  try to update the gaming functionality 
[react-native-fbsdk-next+7.3.3.patch.txt](https://github.com/thebergamo/react-native-fbsdk-next/files/8378576/react-native-fbsdk-next%2B7.3.3.patch.txt)

mikehardy:
[...]

PR #213, opened 2022-03-04
fix(types): allow null value for AppEventsLogger.setUserID

danilobuerger:
AppEventsLogger.clearUserID is deprecated and suggests using "setUserID(null)"
However, setUserID is currently not typed to allow null values.

Test Plan:

no test plan

### DIFF
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -290,7 +290,7 @@ export default {
 
   /**
    * Sets a custom user ID to associate with all app events.
-   * The userID is persisted until it is cleared by clearUserID method.
+   * The userID is persisted until this method is called again with a null userId
    */
   setUserID(userID: string | null) {
     AppEventsLogger.setUserID(userID);
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -292,7 +292,7 @@ export default {
    * Sets a custom user ID to associate with all app events.
    * The userID is persisted until it is cleared by clearUserID method.
    */
-  setUserID(userID: string) {
+  setUserID(userID: string | null) {
     AppEventsLogger.setUserID(userID);
   },
 


src/FBAppEventsLogger.ts
```diff
@@ -292,7 +292,7 @@ export default {
    * Sets a custom user ID to associate with all app events.
    * The userID is persisted until it is cleared by clearUserID method.
```
mikehardy, line 293:
```suggestion
   * The userID is persisted until this method is called again with a null userId
```

mikehardy:
```suggestion
   * The userID is persisted unless it is cleared by calling this method again with a null userID
```

mikehardy (approved):
Oh of course! Should have noticed that when I deprecated the clearUserID method during the forward-port - thanks for catching this. I updated the method doc to match, looks +1

danilobuerger:
@mikehardy Thanks for merging it so quickly. Although I just noticed that calling setUserID(null) now breaks Android. On the android side it doesn't like null but wants clearUserID. Its a shame that iOS and Android are out of sync like that.

mikehardy:
@danilobuerger are you 100% on that? Because I thought they both made that change 

In the current version they internally delegate clearUserId to setUserId(null):
https://github.com/facebook/facebook-android-sdk/blob/ec96d8a0d3fffa3176ef985819650a68e1ba84bb/facebook-core/src/main/java/com/facebook/appevents/AppEventsLogger.kt#L504-L506

and the internal implementation of setUserId explicitly has the id string arg set as nullable
https://github.com/facebook/facebook-android-sdk/blob/ec96d8a0d3fffa3176ef985819650a68e1ba84bb/facebook-core/src/main/java/com/facebook/appevents/AnalyticsUserIDStore.kt#L46

So at least with the current release I expect it will work. However, "the current release" is v13 and we're still pulling in facebook-android-sdk v12 until #208 

on their v12.2 branch it appears to work as well though? https://github.com/facebook/facebook-android-sdk/blob/a6e982175b258fbd99ec8313bbe052f1efd90185/facebook-core/src/main/java/com/facebook/appevents/AppEventsLogger.kt#L498-L500

Except one thing I do see is that setUserId() method is marked with not-null in the Kotlin! So perhaps that's the problem? What exact problem did you see?

We could maybe log an issue for them to true up their types and deprecate clearUserId similar to facebook-ios-sdk

In the meantime if this is breaking things or requiring platform-specific code that's a bug in my way of thinking. For Android we could make setUserId(null) at the javascript layer call native clearUserId() - a quick PR and I'd merge it if you liked the idea and posted it up

danilobuerger:
> Except one thing I do see is that setUserId() method is marked with not-null in the Kotlin! So perhaps that's the problem? What exact problem did you see?

Yeah, that seems to be the problem. I get an error exactly there that a null value is passed to a non null parameter in kotlin.

> In the meantime if this is breaking things or requiring platform-specific code that's a bug in my way of thinking. For Android we could make setUserId(null) at the javascript layer call native clearUserId() - a quick PR and I'd merge it if you liked the idea and posted it up

Yup, will submit a PR later today

mikehardy:
Awesome! Thanks in advance. I'll chase down facebook-android-sdk in parallel with an issue over there

mikehardy:
Upstream already merged the fix sweet! It'll be in whatever is after 13.0.0 whenever that is, and whenever we can access it via forward-porting here :-)

PR #214, opened 2022-03-04
fix(android, setUserId): temporary android workaround for using setUserID(null)

danilobuerger:
Follow up to #213


mikehardy (approved):
Cheesy *and* effective ;-). Thanks!

PR #221, opened 2022-04-04
feat(sdks): Update to v13 of native SDKs

isidoro98:
This is a PR for updating the library to the latest FB sdk as raised on #208 

Facebook post introducing the new version [here](https://developers.facebook.com/blog/post/2022/02/24/introducing-facebook-platform-sdk-13-related-updates/)

# iOS Changes

Please refer to the official changelog [here](https://github.com/facebook/facebook-ios-sdk/blob/main/CHANGELOG.md), main things to note are:

```
The minimum supported version of iOS is now 11.0 for all frameworks. 
The minimum supported version of tvOS is now 11.0 for all frameworks. 
The XCFramework binaries are now built with Xcode 13 so Xcode 12 is no longer supported.
```

```
Starting with the v13.0 release, client tokens must be included in apps for Graph API calls. 
An exception is now raised when running apps without a client token.
```

### Breaking changes

None

# Android Changes

Please refer to the official changelog [here](https://github.com/facebook/facebook-android-sdk/blob/main/CHANGELOG.md), main things to note are:

```
Set Java source and target compatibility to 1.8 (Java 8). 
All apps that integrate with Facebook Android SDK should also set source and compatibility to 1.8 or above.
```

```
GMS AD_ID Permission is added to the SDK by default for requesting the advertising ID
```

```
Removed the support for tokenless requests. 
Now all Graph API requests must be associated with an access token and the app needs to have a client token to construct the access token without a logged in user.
```

### Breaking changes

On android the deprecated ShareLink parameters `contentDescription`, `contentTitle` and `imageUrl` have now been removed.

Regarding the testing plan I would appreciate suggestions as well.

### DIFF
--- a/RNFBSDKExample/ios/RNFBSDKExample.xcodeproj/project.pbxproj
+++ b/RNFBSDKExample/ios/RNFBSDKExample.xcodeproj/project.pbxproj
@@ -11,9 +11,9 @@
 		13B07FBC1A68108700A75B9A /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB01A68108700A75B9A /* AppDelegate.m */; };
 		13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		13B07FC11A68108700A75B9A /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
-		6F42E5EA98685FBE727AD68D /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 5ABCB82D3C55B520AED9AD70 /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a */; };
 		81AB9BB82411601600AC10FF /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */; };
-		D23CBF841BDE50868ED73973 /* libPods-RNFBSDKExample.a in Frameworks */ = {isa = PBXBuildFile; fileRef = C500945E9AE0ABF95A77D925 /* libPods-RNFBSDKExample.a */; };
+		C9654CFD18664A5E3CB7772A /* libPods-RNFBSDKExample.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 9FBB99C19935EAEEB6EBC319 /* libPods-RNFBSDKExample.a */; };
+		D014B9EF2FA3510E228B1412 /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 4B41EF6F0EB71524760B1B5F /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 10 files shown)


RNFBSDKExample/ios/RNFBSDKExample.xcodeproj/project.pbxproj
```diff
@@ -11,9 +11,9 @@
 		13B07FBC1A68108700A75B9A /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB01A68108700A75B9A /* AppDelegate.m */; };
 		13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		13B07FC11A68108700A75B9A /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
-		6F42E5EA98685FBE727AD68D /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 5ABCB82D3C55B520AED9AD70 /* libPods-RNFBSDKExample-RNFBSDKExampleTests.a */; };
```
mikehardy, line 14: Did you update project.pbxproj manually, or is this the result of https://github.com/thebergamo/react-native-fbsdk-next/blob/master/refresh-example.sh ?

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
```diff
@@ -168,10 +168,6 @@ public static GameRequestContent buildGameRequestContent(ReadableMap gameRequest
     public static ShareLinkContent buildShareLinkContent(ReadableMap shareLinkContent) {
         ShareLinkContent.Builder contentBuilder = new ShareLinkContent.Builder();
         contentBuilder.setContentUrl(Uri.parse(shareLinkContent.getString("contentUrl")));
-        String url = getValueOrNull(shareLinkContent, "imageUrl");
-        contentBuilder.setImageUrl(url != null ? Uri.parse(url) : null);
-        contentBuilder.setContentDescription(getValueOrNull(shareLinkContent, "contentDescription"));
-        contentBuilder.setContentTitle(getValueOrNull(shareLinkContent, "contentTitle"));
```
mikehardy, line 174: These are still referenced in our docs though https://github.com/thebergamo/react-native-fbsdk-next#share-dialogs - how should people adapt?

mikehardy (changes requested):
Nice! Thank you so much for posting this

Have you test-integrated it in to an app? You can do so by creating a patch-package patch from inside the example with your changes, then using that in a work project - if the example works and your use cases work then I think technically it's good.

With breaking changes the biggest thing is documentation either directly in the changelog (my favorite) or in a migration guide linked from the changelog, and updating the docs

I left a comment related to the share dialog, it looks like the docs need an update?
After that I think the breaking changes to call out are:
- ios min version moved to 11
- you must use Xcode 13
- share dialog parameters changed as mentioned
- you must use configure and use a client token to use Graph API calls
- GMS AD_ID Permission is added to the SDK by default, use Android Manifest merge tools to remove it if unwanted

Anything else? That was what I got from your PR description. That's short enough I can include it as a breaking changes entry on the merge so it shows up in the changelog

RNFBSDKExample/ios/RNFBSDKExample.xcodeproj/project.pbxproj
isidoro98: A result of the script

android/src/main/java/com/facebook/reactnative/androidsdk/Utility.java
isidoro98: Looks like those 3 parameters that were already marked as deprecated have now been deleted, not much people can do beside stop using them right?

mikehardy: Okay - can you add a commit with changes to this page https://github.com/thebergamo/react-native-fbsdk-next#share-dialogs to reflect what is possible now?

Other than that this is probably good to go then, I only use login myself so I'm not a great test, although the example app does at least attempt a share. If there are problems others can propose fix ups I think, it's open source...

isidoro98: Awesome, just updated it. Thank you so much for your help Mike

mikehardy (approved):
Okay - I think this looks good.

For anyone reading this now we need testers! That is what will gate this from release. I would like to hear one other person say they have test-integrated it and it works for them, ideally that uses some of the other modules (login and events seem to be covered).

I can do this but I am very short on time so it might be a few days or more. Worst case, if it sits for a week I'll just release it and if there are issues, there are issues and we do x.0.1 release

Either way, fantastic work @isidoro98 thank you!

mikehardy:
I posted the patch-package that contains this change on the related issue https://github.com/thebergamo/react-native-fbsdk-next/issues/208#issuecomment-1092935480

abumalick:
Thanks for working on this @isidoro98 . I tested the build on expo managed app via the plugin from the repository. The  expo-dev-client build passed, on android it seems to work. On iOS, it crashes when I launch the app with this error:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Starting with v13 of the SDK, a client token must be embedded in your client code before making Graph API calls.
Visit https://developers.facebook.com/apps/738908110438103/settings/advanced/ to find your client token for this app.
Add a key named FacebookClientToken to your Info.plist, and add your client token as its value.
Visit https://developers.facebook.com/docs/ios/getting-started#configure-your-project for more information.'
```


The facebook client token is not mandatory for iOS. We should add this to the documentation right? This was not required before and is also a breaking change for iOS.

In the expo plugin (plugin folder at the root of the repo) we could make the build throw an error if clientToken is not passed also (But I understand if you prefer to do this on another pull request)

I am adding the client token and give more feedback inchallah.

mikehardy:
Needing a client access token is definitely a breaking change, this will be a breaking change release as noted in my proposed changelog notes. If the Expo plugin needs an update for this we'll need someone with experience to handle it, you can do that right now in parallel on a separate PR, no need to wait. Optionally see if there is an fbsdk client access token for ios in app.json and do the needful if it exists? it will work on v12 already

abumalick:
I don't see that the client token is necessary in the readme for native instructions. Shouldn't we add something there?

mikehardy:
> I don't see that the client token is necessary in the readme for native instructions. Shouldn't we add something there?

Yes! Good catch. It was optional before but now is definitely mandatory. That's a README update we should add in this PR if you can @isidoro98 or we can do a follow-up

abumalick:
> If the Expo plugin needs an update for this we'll need someone with experience to handle it, you can do that right now in parallel on a separate PR, no need to wait. 

I am a bit busy these days, I will try to work on this this weekend or in the beginning of the week inchallah

isidoro98:
> > I don't see that the client token is necessary in the readme for native instructions. Shouldn't we add something there?
> 
> Yes! Good catch. It was optional before but now is definitely mandatory. That's a README update we should add in this PR if you can @isidoro98 or we can do a follow-up

The readme is already linking to facebook's getting started Guide of both iOS and android, where it explains the configuration step by step, including adding the client token. The README itself doesn't get into detail regarding how to configure the app.

Can be a troubleshoot step, what do you guys think?

mikehardy:
> The readme is already linking to facebook's getting started Guide of both iOS and android, where it explains the configuration step by step, including adding the client token. The README itself doesn't get into detail regarding how to configure the app.
> 
> Can be a troubleshoot step, what do you guys think?


Ahh, good point. Maybe a "I just upgraded to v13 and it says I need an access token"
Since we defer to the upstream getting started guide, I think that (for new users) plus the changelog (for upgrades) plus a troubleshoot step would handle 110% of all cases ;-)

mikehardy:
I'm going to merge this - the Expo plugin update I will peel off to a new issue, for people that need the Expo plugin to work they'll have to stick on v12 until it's updated

PR #222, opened 2022-04-05
fix(android): get current profile always returns null in the first time

### LINKED ISSUE #73: [Android] Profile.getCurrentProfile() always returns null in the first time

maykonmichel:
Fixes #73

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
@@ -18,24 +18,33 @@ public class FBProfileModule extends ReactContextBaseJavaModule {
     public static final String NAME = "FBProfile";
 
     public FBProfileModule(ReactApplicationContext reactContext) {
-      super(reactContext);
+        super(reactContext);
     }
 
     @NonNull
     @Override
     public String getName() {
-    return NAME;
-  }
+        return NAME;
+    }
 
     /**
-    * Get the current logged profile.
-    * @param callback Use callback to pass the current logged profile back to JS.
-    */
+     * Get the current logged profile.
+     *
+     * @param callback Use callback to pass the current logged profile back to JS.
+     */
     @ReactMethod
     public void getCurrentProfile(Callback callback) {
         //Return the profile object as a ReactMap.
-        callback.invoke(Profile.getCurrentProfile() == null
-                ? null
-                : Utility.profileToReactMap(Profile.getCurrentProfile()));
-  }
+        if (Profile.getCurrentProfile() == null) {
+            mProfileTracker = new ProfileTracker() {
+                @Override
+                protected void onCurrentProfileChanged(Profile oldProfile, Profile currentProfile) {
+                    callback.invoke(Utility.profileToReactMap(currentProfile));
+                    mProfileTracker.stopTracking();
+                }
+            };
+        } else {
+            callback.invoke(Utility.profileToReactMap(Profile.getCurrentProfile()));
+        }
+    }
 }


android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
```diff
@@ -18,24 +18,33 @@
...
+     */
     @ReactMethod
     public void getCurrentProfile(Callback callback) {
         //Return the profile object as a ReactMap.
-        callback.invoke(Profile.getCurrentProfile() == null
-                ? null
-                : Utility.profileToReactMap(Profile.getCurrentProfile()));
-  }
+        if (Profile.getCurrentProfile() == null) {
+            mProfileTracker = new ProfileTracker() {
```
mikehardy, line 39: it doesn't appear that you have actually tested this at all? I say that because this won't compile, and that's a worrying sign - https://github.com/thebergamo/react-native-fbsdk-next/runs/5838884802?check_suite_focus=true#step:9:395

android/src/main/java/com/facebook/reactnative/androidsdk/FBProfileModule.java
```diff
@@ -18,24 +18,33 @@
...
+                @Override
+                protected void onCurrentProfileChanged(Profile oldProfile, Profile currentProfile) {
+                    callback.invoke(Utility.profileToReactMap(currentProfile));
+                    mProfileTracker.stopTracking();
+                }
+            };
```
mikehardy, line 45:
```suggestion
            AtomicBoolean hasInvoked = new AtomicBoolean(false);
            new ProfileTracker() {
                @Override
                protected void onCurrentProfileChanged(Profile oldProfile, Profile currentProfile) {
                    if (hasInvoked.compareAndSet(false, true)) {
                      callback.invoke(Utility.profileToReactMap(currentProfile));
                    }
                    mProfileTracker.stopTracking();
                }
            };
            try { Thread.currentThread().sleep(3000); } catch (InterruptedException e) { /* do nothing */ }
            hasInvoked.compareAndSet(false, true) {
              callback.invoke(null);
            }
```


Or similar, but I haven't compiled this either, as mentioned in the related issue it's not a use case of mine so I don't have time to pull it and test it, but something like this is needed I think, otherwise I think there will be issues in the future related to promises that don't resolve sometimes...

What do you think?

mikehardy (changes requested):
This looks like a good start to me - thanks for posting the PR.

I'm a bit concerned about testing since it doesn't compile, and I'm also worried about infinite hangs, I left some details comments on the same

PR #223, opened 2022-04-07
fix(jest): add LoginManager.logOut + AccessToken mocks

Andarius:
Fix issue where `jest.spyOn(LoginManager, 'logOut')` would raise `Cannot spy the logOut property because it is not a function; undefined given instead`

### DIFF
--- a/jest/mocks/index.js
+++ b/jest/mocks/index.js
@@ -48,6 +48,10 @@ export const mockAppEventParams = {
 };
 
 export default {
+  AccessToken: {
+    refreshCurrentAccessTokenAsync: jest.fn(),
+    getCurrentAccessToken: jest.fn(),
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Nice! Thank you
Don't worry about the conventional commit CI job, that's a quirk of the way GitHub does merges and is just a warning to maintainers at this point to be careful at the merge

mikehardy (approved):
previously looked good to me, now even better ;-). You want me to hang on while you check through other stuff or you want me to merge as soon as CI goes green?

Andarius:
Hey @mikehardy, I think I covered everything I needed, all my tests are passing now. I made a patch while waiting for a release so up to you :)

PR #249, opened 2022-05-09
fix(expo, ios): add client token to info plist (fixes #228)

### LINKED ISSUE #228: Expo config plugin needs v13 update for client access token
Thanks for working on this @isidoro98 . I tested the build on expo managed app via the plugin from the repository. The  expo-dev-client build passed, on android it seems to work. On iOS, it crashes when I launch the app with this error:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Starting with v13 of the SDK, a client token must be embedded in your client code before making Graph API calls.
Visit https://developers.facebook.com/apps/738908110438103/settings/advanced/ to find your client token for this app.
Add a key named FacebookClientToken to your Info.plist, and add your client token as its value.
Visit https://developers.facebook.com/docs/ios/getting-started#configure-your-project for more information.'
```

The facebook client token is not mandatory for iOS. We should add this to the documentation right? This was not required before and is also a breaking change for iOS.

In the expo plugin (plugin folder at the root of the repo) we could make the build throw an error if clientToken is not passed also (But I understand if you prefer to do this on another pull request)

I am adding the client token and give more feedback inchallah.

_Originally posted by @abumalick in https://github.com/thebergamo/react-native-fbsdk-next/issues/221#issuecomment-1094902670_

thmsobrmlr:
I'm in the process of migrating from `expo-facebook` to this library. Thereby I ran into the issue mentioned in #228, where the new SDK requires the facebook client token to be set in the info plist. This PR adds the client token to the expo config plugin, so that the workaround does not need to be used.


mikehardy (approved):
This looks great to me! Thanks for posting this up

PR #250, opened 2022-05-12
fix(expo, ios): pass through config for ios user tracking permission

thmsobrmlr:
When trying to submit an update via App Store Connect, I was greeted with following issue:

Screenshot 2022-05-12 at 09 39 26

The cause for this is that the expo plugin config variable `iosUserTrackingPermission` is not passed through. Thus the `withUserTrackingPermission` sub-function will always add the default user tracking permission string. This pull request fixes this.

Test Plan:
I linked this package locally and ran `expo prebuild` once with the change and once without, while checking `ios/MyProject/Info.plist` for `NSUserTrackingUsageDescription`. Without change, it is present. With change, it is gone.

--

I have also added tests for `withUserTrackingPermission`. Please let me know if you'd like to keep them. The tests would not have found the issue, since they only test the `withUserTrackingPermission` function.

It would be possible to either convert the existing tests or to add new tests in a more e2e-style by invoking the main plugin function `withFacebook` and then checking the output. What is your opinion on this?


mikehardy (approved):
This looks great to me - thanks for the tests especially, they look like full coverage, a rare thing - I don't use Expo personally but I'm getting more familiar as config plugins ooze out into all my repos :-). This seems like it should do the trick and I really appreciate the help

thmsobrmlr:
Thanks for the kind words! Let me know when you need some help / feedback regarding Expo in one of your repos. I have some exposure to the plugins through my own project.

PR #251, opened 2022-05-12
refactor(formatting): expand prettier scope to include jest and plugin files

thmsobrmlr:
This PR:
- Removes the [require pragma](https://prettier.io/docs/en/options.html#require-pragma) option from the prettier config. Only one file needed formatting, so fully transitioning to prettier here should not hurt.
- Includes the jest and plugin folders in the prettier config. Let me know what you think.

### DIFF
--- a/jest/mocks/index.js
+++ b/jest/mocks/index.js
@@ -1,7 +1,6 @@
 /* eslint-env jest */
 const LoginButton = require('../../lib/commonjs/FBLoginButton').default;
 
-
 export const mockAppEvents = {
   AchievedLevel: 'fb_mobile_level_achieved',
   AdClick: 'AdClick',
--- a/package.json
+++ b/package.json
@@ -121,7 +121,6 @@
     "typescript": "^4.5.5"
   },
   "prettier": {
-    "requirePragma": true,
     "singleQuote": true,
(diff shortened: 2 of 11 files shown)


mikehardy (changes requested):
Nice! For formatting, philosophically I like:
- fully automated formatting in all the places so there's no time wasted debating
- as simple a formatting setup as possible, so there's less time maintaining
- commits/PRs that do formatting style changes all by themeselves

You are ticking all the boxes here, so this is perfect :-)

I would merge immediately except -
- took a conflict from your functional change on the plugin (of course...)
- some small formatting issue in CI:

```

$ yarn validate:prettier && yarn validate:eslint && yarn jest
$ prettier "{src,jest,plugin/src}/**/*.{js,ts,tsx}" --check
Checking formatting...
[warn] jest/mocks/index.js
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

If you can resolve those two things we are in business, with even prettier code...

thmsobrmlr:
Thanks for the feedback! I rebased the pull request and included the missing file. Now it should be good to go.

mikehardy (approved):
Indeed! Clean and :sparkles: now, thanks for this

PR #256, opened 2022-05-18
fix: fix argument types in setDataProcessingOptions native java method

oscb:
Thank you for sending the PR! We appreciate you spending the time to work on these changes.
Help us understand your motivation by explaining why you decided to make this change.

If this PR fixes an issue, type "Fixes #issueNumber" to automatically close the issue when the PR is merged.

Test Plan:

Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
@@ -13,7 +13,10 @@ import androidx.annotation.Nullable;
 import com.facebook.FacebookSdk;
 import com.facebook.react.bridge.BaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.module.annotations.ReactModule;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This is a {@link NativeModule} that allows JS to use SDK settings in Facebook Android SDK.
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
```diff
@@ -30,24 +33,17 @@ public String getName() {
...
      * @param options list of the options
      * @param country code of the country
      * @param state code of the state
      */
     @ReactMethod
-    public static void setDataProcessingOptionsExtra(@Nullable String[] options, int country, int state) {
-        FacebookSdk.setDataProcessingOptions(options, country, state);
+    public void setDataProcessingOptions(ReadableArray options, int country, int state) {
+        List<String> opts = Utility.reactArrayToStringList(options);
+        String[] optsArray = opts.toArray(new String[0]);
```
mikehardy, line 45: I thought this construct was odd then when I went to check it, it appears your stackoverflow skills are the same as mine - https://stackoverflow.com/a/4042464/9910298 - or you just do this often enough you already knew the optimal conversion idiom ;-)

mikehardy (approved):
Looks great - thanks for patching this up. Just pending CI now but I am curious - does it work on iOS? I guess if the NSArray and NSStringArray types are compatible that one just works already?

android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
oscb: Java is not my strongest language so I also find it odd, but When in Rome... :P

oscb:
> I guess if the NSArray and NSStringArray types are compatible that one just works already?
Yea, I believe NSStringArray is a typedef to NSArray<NSString *> in the React Native code.

PR #285, opened 2022-07-01
fix(expo, ios)!: iosUserTrackingPermission no longer added by default in Expo

### LINKED ISSUE #272: NSUserTrackingUsageDescription included by default on Expo
## 🐛 Bug Report
`react-native-fbsdk-next` is including `NSUserTrackingUsageDescription` by default in `infoPlist`.

## To Reproduce
Just install this library as dependency in your Expo project, and you'll see `NSUserTrackingUsageDescription` added by default in your `infoPlist` file, after building with `eas build`.

## Expected Behavior
Not to be included by default (in the `infoPlist` file)

foloinfo:
Fixes #272

Test Plan: `yarn test`

Please give any feedback whether it is better to have default `NSUserTrackingUsageDescription` or make it optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -721,7 +721,7 @@ The plugin provides props for extra customization. Every time you change the pro
 - `appID` (_string_): Facebook Application ID.
 - `displayName` (_string_): Application Name.
 - `clientToken` (_string_): Client Token.
-- `iosUserTrackingPermission` (_string_): iOS User Tracking Permission. Defaults `This identifier will be used to deliver personalized ads to you.`.
+- `iosUserTrackingPermission` (_string_): iOS User Tracking Permission.
 - `advertiserIDCollectionEnabled` (_boolean_): Enable advertiser ID collection. Default `false`.
 - `autoLogAppEventsEnabled` (_boolean_): Default `false`.
 - `isAutoInitEnabled` (_boolean_): Default `false`.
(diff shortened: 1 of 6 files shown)


mikehardy (approved):
looks good to me, and, actually looks backwards compatible, if previous users were using `false` to avoid the permission, they will still avoid the permission...

mikehardy:
Not sure how this passed your builds locally while you were testing this - you were testing it locally pre-submission right?

Appears to fail during initial build with type errors

```

ℹ Generating type definitions with tsc
plugin/src/__tests__/withUserTrackingPermission-test.ts:8:10 - error TS2322: Type 'boolean' is not assignable to type 'string'.
8         {iosUserTrackingPermission: false},
         ~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in plugin/src/__tests__/withUserTrackingPermission-test.ts:8
✖ Failed to build definition files.
bob build
```

foloinfo:
Oops! I only run `yarn test locally, I'take a look.

foloinfo:
It was failing with `yarn validate:ts` (I'm not used to ts), I checked with `npm run clean:plugin && npm run build:plugin` and it seems like working, so I hope it passes now.

I also realized if I removed bool type from `string | bool`, then the current user might be forced to change their settings (which will do nothing), so I kept the interface same, rather than removing it.

@mikehardy Could you re-check the changes?

mikehardy (approved):
LGTM

mikehardy:
I merged it as a breaking change with as clear a message as I could think of telling who was affected and who could ignore, and if affected what exactly it is. Those never bother me as a developer, so, should be good to go

PR #310, opened 2022-09-12
(doc) Add fix for login with Facebook app on iOS

chuchuva:


mikehardy (changes requested):
Hey :wave: - thanks for helping fix up the docs, upstream has moved to Swift and isn't very clear, so helping people through this part will be nice

I'd prefer for the steps here: https://github.com/thebergamo/react-native-fbsdk-next#32-ios

to be updated to add the code you put here in troubleshooting - it's actually a basic step everyone needs to do, and the FBSDK upstream docs have moved to Swift which might be confusing.

We already include a snippet at the link I pasted in that shows how to do the initialization, putting a snippet for the openURL implementation right there would be best in my opinion

chuchuva:
@mikehardy Done! Thanks for reviewing.

README.md
```diff
@@ -133,6 +133,18 @@ Follow ***steps 2, 3 and 4*** in the [Getting Started Guide](https://developers.
       ```
    3. After this step, if you run into this `build` issue: `Undefined symbols for architecture x86_64:`, 
    then you need to create a new file `File.swift` on your project folder. After doing this, you will get a prompt from `Xcode` asking if you would like to create a `Bridging Header`. Click accept.
+   4. Add the following code to AppDelegate.m:
```
mikehardy, line 136:
```suggestion
   4. From the facebook-ios-sdk docs steps 1-3, but in Objective-C since they have moved to Swift for their examples - make something like the following code is in AppDelegate.m:
```

README.md
```diff
@@ -133,6 +133,18 @@ Follow ***steps 2, 3 and 4*** in the [Getting Started Guide](https://developers.
...
+      - (BOOL)application:(UIApplication *)app
+                  openURL:(NSURL *)url
+                  options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options
+      {
+        return [[FBSDKApplicationDelegate sharedInstance]application:app
+                                                            openURL:url
+                                                            options:options];
+      }
+      ```
+      Without this code login might not work if Facebook app is installed, see https://github.com/thebergamo/react-native-fbsdk-next/issues/59#issuecomment-1038149447
```
mikehardy, line 147:
```suggestion
      Without this code login might not work if Facebook app is installed, see https://github.com/thebergamo/react-native-fbsdk-next/issues/59#issuecomment-1038149447 - if you are also using react-native deep-linking you may need have multiple entries in this openURL method, as detailed in the next section
```

mikehardy (approved):
This looks good and should clear up some confusion I think. Thanks a bunch!

PR #313, opened 2022-09-29
feat(sdks)!: new v15 SDKs, iOS requires minimum iOS/tvOS of 12 now

mikehardy:
Update to new native SDKs

updated example to current react-native as well

Test Plan:

Run the example locally before / after, everything seems fine


mikehardy:
Note, blocked on an Android 12+ issue here, android build will be successful when facebook-android-sdk 15.0.1 is released: https://github.com/facebook/facebook-android-sdk/issues/1114

PR #315, opened 2022-09-30
docs(ios, AEM): AEM setup requires a reporter parameter in API call

Gp2mv3:
I corrected the configuration of FBAEMReporter for iOS as it needs a reporter parameter.


mikehardy (approved):
Thanks for posting this!

PR #322, opened 2022-10-03
docs(ios): Update docs related to the last fix

HorbachAndrii:
Update docs related to the [last fix](https://github.com/thebergamo/react-native-fbsdk-next/pull/320)


mikehardy (approved):
I think you are right and this looks good to me!

PR #325, opened 2022-10-07
docs(android): add tip about Key Hash generation

### LINKED ISSUE #321: [Android] Fix for wrong Development Key Hash creation command
# Problem

When I tried to use this package in my Android app, I've encountered an issue while configuring it to work with my Facebook app. To be more specific, the Key Hash generation command in the [Running Sample Apps](https://developers.facebook.com/docs/android/getting-started/#samples) part of Facebook's `Getting Started Guide` linked under `Installation` -> `3. Configure projects` -> [`3.1 Android`](https://github.com/thebergamo/react-native-fbsdk-next#31-android) in the project's README.

I've encountered this error after using the command from Facebook's guide:

```
SERVER_ERROR: [code] 1349195 [message]: The key hash does not match any stored key hashes. Go to https://developers.facebook.com/docs/facebook-login/android for more information. [extra]
```

# Solution

Well, this is caused by the fact that the RN app, doesn't get it's `debug.keystore` from `~/.android/debug.keystore` (where Facebook's command puts it). Facebook's command is simply not correct for a RN project (both Expo and React Native CLI). This is of course the case because Facebook's guide was written with native Android development in mind.

Nonetheless I've figured out that using the `debug.keystore` inside `./android/app` (from the root of a RN app), resolves the problem. Therefore I've though up this process:

1. Get in to the root dir of your app:

```bash
cd my-app
```

2. If you're using Expo, you'll of course need to run `npx expo prebuild` first.
3. Generate the Development Key Hash with the fixed command (Linux / MacOS):

```bash
keytool -exportcert -alias androiddebugkey -keystore android/app/debug.keystore | openssl sha1 -binary | openssl base64
```

4. Enter `android` or the password you've changed it to (if it doesn't ask you for a password, you either don't have all the needed utilities installed / the path is wrong / there's something wrong with the command).
5. Copy & paste the Development Key Hash in to your Facebook app.
6. Update algorithm as SHA1 is a security risk (Linux / MacOS):

```bash
keytool -importkeystore -srckeystore android/app/debug.keystore -destkeystore android/app/debug.keystore -deststoretype pkcs12
```

From my testing it looks like the app (both Expo and React Native CLI) doesn't even need to be reloaded, let alone rebuilt after running the Key Hash generation command.

# P.S.

**I think this (with some polish - like adding equivalent Windows commands and so on) would be a worthy addition to the project's README!**

ErykDarnowski:
Fixes #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ protected List<ReactPackage> getPackages() {
 
 #### 3.1 Android
 
-Before you can run the project, follow the [Getting Started Guide](https://developers.facebook.com/docs/android/getting-started/) for Facebook Android SDK to set up a Facebook app. You can skip the build.gradle changes since that's taken care of by the rnpm link step above, but **make sure** you follow the rest of the steps such as updating `strings.xml` and `AndroidManifest.xml`.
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Looks good to me with this wording, thanks a lot for working through that :-)

PR #343, opened 2022-11-08
Update README to better clarify config plugin options

bradjones1:
The `scheme` option is also not listed and is required for proper operation, e.g. on iOS with Facebook app installed.

Docs-only change.


mikehardy (approved):
I'll take your word for it, thanks a  bunch for the configuration documentation help, stops problems before they start...

PR #370, opened 2023-01-04
fix: type definition of AppEvents and AppEventParams in FBAppEventsLogger.ts

w00ing:
The compiled version of present `FBAppEventsLogger.d.ts` looks like this:
```
...
    /**
     * Predefined event names for logging events common to many apps.
     */
    AppEvents: any;
    /**
     *  Predefined event name parameters for common additional information to accompany events logged through the `logEvent`.
     */
    AppEventParams: any;
};
...
```

As you can see, `AppEvents` and `AppEventParams` are typed as any. This PR attempts to fix this issue.

### DIFF
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -134,7 +134,10 @@ export type AppEventParam = {
   ValueYes: string;
 };
 
-const {AppEvents, AppEventParams} = AppEventsLogger?.getConstants() || {};
+const {AppEvents, AppEventParams} = (AppEventsLogger?.getConstants() || {}) as {
+  AppEvents: AppEvent;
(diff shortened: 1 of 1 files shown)


mikehardy:
sorry for the inattention, always hard to find time.
Reopening for consideration

mikehardy:
Maybe superceded by #422 - I *think* that is a cleaner declaration, but @w00ing what do you think?

w00ing:
> Maybe superceded by #422 - I _think_ that is a cleaner declaration, but @w00ing what do you think?

Agree, #422 approach would be better. Thank you for your work!

mikehardy:
Thanks for checking that one out and giving it some thought, much appreciated. I'll merge that one and get the fix out for you and everyone :-)

PR #376, opened 2023-01-10
fix(readme): iosUserTrackingPermission: string | false

Edmx747:
Update readme for iosUserTrackingPermission config. This optionnal value can be a string or false.


mikehardy:
you are correct! Thank you

mikehardy (approved):

PR #409, opened 2023-04-08
fix(expo): Update expo config plugin dependencies

### LINKED ISSUE #404: Version bump @expo/config-plugins to 6.0.0?

DoctorJohn:
Expo doctor (`npx expo-doctor`) warns about this package using an outdated version of expo config plugins. This PR aligns the dependencies used by this packages config plugin with [the latest expo recommendations](https://docs.expo.dev/home/config-plugins/development-and-debugging/). They now recommend having expo as an optional peer dependency. It also says:

> For simple config plugins that depend on core, stable APIs, such as a plugin that only modifies AndroidManifest.xml or Info.plist, you can use a loose dependency [...]

I verified that's the case.

Fixes #404
(only putting that here so said issue gets automatically closed).

Test Plan:

I ran the unit tests for the config plugin (`yarn test`) which verify the modifications made by the plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,11 +84,16 @@
     "plugin/build/"
   ],
   "dependencies": {
-    "@expo/config-plugins": "^5.0.1",
     "xml2js": "^0.4.23"
   },
   "peerDependencies": {
+    "expo": ">=47.0.0",
     "react-native": ">=0.63.3"
+  },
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
LGTM - thanks @DoctorJohn - much appreciated

PR #418, opened 2023-04-20
Fix to make it work with Xcode 14.3

chimame:
## Problem

I fixed two issues that were causing the sample program to not work.

1. Fixed a bug where pod install couldn't be executed properly.
2. Fixed a bug where the program wouldn't work on Xcode 14.3.

## Solution

1. When executing `pod install`, the following error occurred.

```
[!] CocoaPods could not find compatible versions for pod "FBSDKLoginKit":
  In snapshot (Podfile.lock):
    FBSDKLoginKit (= 15.0.0, ~> 15.0.0)

  In Podfile:
    react-native-fbsdk-next (from `../node_modules/react-native-fbsdk-next`) was resolved to 10.1.0, which depends on
      react-native-fbsdk-next/Login (= 10.1.0) was resolved to 10.1.0, which depends on
        FBSDKLoginKit (~> 14.1.0)


You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * changed the constraints of dependency `FBSDKLoginKit` inside your development pod `react-native-fbsdk-next`.
   You should run `pod update FBSDKLoginKit` to apply changes you've made.
```

Therefore, I updated the version of `react-native-fbsdk-next` to the latest one, since it was managed by `yarn.lock` and was outdated.

2. Building React Native failed on Xcode 14.3. Therefore, we upgraded to the latest version, 0.70.8, which fixed the issue.

https://github.com/facebook/react-native/issues/36739

### DIFF
--- a/RNFBSDKExample/ios/Podfile.lock
+++ b/RNFBSDKExample/ios/Podfile.lock
@@ -87,7 +87,7 @@ PODS:
     - FlipperKit/FlipperKitNetworkPlugin
   - fmt (6.2.1)
   - glog (0.3.5)
-  - hermes-engine (0.70.1)
+  - hermes-engine (0.70.8)
   - libevent (2.1.12)
   - OpenSSL-Universal (1.1.1100)
   - RCT-Folly (2021.07.22.00):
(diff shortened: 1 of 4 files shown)


mikehardy:
The example app is currently managed by [refresh-example.sh](https://github.com/thebergamo/react-native-fbsdk-next/blob/master/refresh-example.sh) - so the best thing to do is to update that (as needed) and make a commit with the updates, then run it and commit the generated example, then propose that as a PR

The changes you proposed are necessary - just need to do them in a slightly different way. I'm working through that now as I also had the same "can't build the example..." issue today when looking to merge a couple PRs 😆

PR #422, opened 2023-04-28
feat(EventLogger): add type for AppEvents and AppEventsParams

alefemafra:
Adding type to AppEvents and AppEventsParams to be able to see what types are available to be used following AppEventsLogger.AppEvents, making it easier when creating a log.

### DIFF
--- a/src/FBAppEventsLogger.ts
+++ b/src/FBAppEventsLogger.ts
@@ -134,7 +134,11 @@ export type AppEventParam = {
   ValueYes: string;
 };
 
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Thanks for your patience - this looks good!

mikehardy:
@alefemafra did you see the preceding work (that staled out, I think 😬 ) here? #370 

Which do you think is better? I think this one is better as it appears to be a cleaner declaration personally

mikehardy (approved):
Thanks for this!

PR #458, opened 2023-09-12
docs(ios): fix 'FBSDKCoreKit/FBSDKCoreKit-swift.h' file not found on case-sensitive filesystems

Romick2005:
This fix the issue for case-sensitive file systems. 
It would have no impact for not case-sensitive user systems.

Build error:

image


mikehardy (approved):
Nice one! That is subtle, in so far as very few people use case-sensitive file systems for macOS but for those that do, this is very important. Thank you!

PR #485, opened 2024-01-20
feat(android): RN-0.73 and Android Gradle Plugin 8.0 Compatibility Fix

dishantwalia:
Test Plan:
https://github.com/react-native-community/discussions-and-proposals/issues/671

React Native 0.73 will depend on Android Gradle Plugin (AGP) 8.x. This will require all the libraries to specify a namespace in their build.gradle file. We added a compatibility layer for libraries that haven't specified a namespace, but please consider updating your libraries nonetheless.

Screenshot 2024-01-20 at 5 46 18 PM

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -18,5 +18,6 @@
     CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 -->
 
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest
+    package="com.facebook.reactnative.androidsdk">
 </manifest>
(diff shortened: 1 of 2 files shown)


android/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,5 @@
     CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 -->
 
-<manifest
-    package="com.facebook.reactnative.androidsdk">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
mikehardy, line 21: this is a non-backwards-compatible change and not the way I've done it in other modules

mikehardy, line 21:
```suggestion
<manifest
    package="com.facebook.reactnative.androidsdk">
```

mikehardy (approved):
this now corresponds to the way I've done it in other modules which will emit a warning because of the AndroidManifest.xml contents but will still work for people on all versions of AGP (5, 6, 7 and 8)

This corresponds to the patch I'm currently using in my work apps for this module and works + is in production / tested

ankitpunchh:
Thank you for this change.

PR #505, opened 2024-03-25
feat: add support for new architecture in compatibility mode

gabrieldonadel:
### Summary 

This PR adds Android and iOS support for the new architecture relying on the compatibility layer. 

#### On Android 
two main changes were required for this to work:

1. Remove `static` from all functions annotated with `@ReactMethod` given that in bridgeless it would throw the following error 

2. Use the event dispatcher to send events from the native side to JS and adding a RCTLoginButtonEvent class

#### On iOS 

Had to create a dedicated class for the RCTFBSDKLoginButton view in order to export the `onChange` view property of type  `RCTBubblingEventBlock` and declare this property on the UIView wrapping the button (`FBSDKLoginButton` is a final class and can't be extended)

### Test Plan:

Locally run `refresh-example` to update to the latest RC and tested Android and iOS on the old and new architectures

https://github.com/thebergamo/react-native-fbsdk-next/assets/11707729/891ab1aa-23c5-4ee0-a5e5-cb1ebee31700

https://github.com/thebergamo/react-native-fbsdk-next/assets/11707729/8fec570b-23b5-49f4-84bd-093ad2c60b57

### DIFF
--- a/android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
+++ b/android/src/main/java/com/facebook/reactnative/androidsdk/FBSettingsModule.java
@@ -51,7 +51,7 @@ public class FBSettingsModule extends BaseJavaModule {
      * [FB SDK Best Practices for GDPR Compliance](https://developers.facebook.com/docs/app-events/gdpr-compliance/)
      */
     @ReactMethod
-    public static void initializeSDK() {
+    public void initializeSDK() {
         FacebookSdk.fullyInitialize();
     }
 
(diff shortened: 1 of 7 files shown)


mikehardy (approved):
Thanks for this @gabrieldonadel ! Looks good on review and appears to work locally
I've done a chunk of work here for QA purposes, to make sure CI *can* go green and to refresh the example, and things appear to work well at least on a backwards-compatibility check

Note that I have *not* initialized a react-native 0.74-rc project with new arch enabled to see if this works in bridgeless which is it's purpose, but I assume you've done so or what's the point 😆

habovh:
@mikehardy I'd like to bring #645 to your attention since the breaking changes have been introduced by this PR that you merged.

PR #506, opened 2024-03-29
bump iOS SDK version to 17.0.0

efstathiosntonas:
`16.3.1` version has issues with Xcode 15.3:

* https://github.com/facebook/facebook-ios-sdk/issues/2353

PR #511, opened 2024-04-10
build(deps): some basic dependency updates

mikehardy:
Each commit is valid on its own

After rebuilding the example I ran android and ios locally and they appeared to work fine

Ci should check the action changes, and everything is ✅  this is good to go


mikehardy:
Github actions runners are having an incident now even though their uptime status board isn't showing it yet.
I believe this PR is good to go but I need CI to go green which means I need Github to notice the problem then fix it then re-run CI. So this is blocked for the moment but should be ready

This will in-turn unblock #506 I believe as well as the merge of the other PRs open

PR #557, opened 2024-09-26
feat(ios): adopt facebook-ios-sdk 17.1.0

### LINKED ISSUE #555: FB SDK version Needs upgrade (from v17.0.3 to v17.1.0)
Hello, Facebook has been in need of an update to a new version for a long time, and without an update, launching ads for events in the application does not work... Also, Facebook constantly reminds you that you need to update the SDK. Unfortunately, the last update here was 5 months ago, and it's not clear if someone is working on an update now...

That is why I am asking for help, what other solutions can there be to integrate Facebook with React Native, so that it is not necessary to use xcode... Maybe some modules or something else...

mikehardy:
- update example to use current-reactive and fuzz-bust through all the issues related
  - new react-native init items like git init
  - yarn2 is the default now, so converted module and example
  - new syntax for yarn2 git adds
  
Investigate the diff between facebook-ios-sdk 17.0.x and 17.1.x, appears fine, and *appears* to work fine in testing

Note: there is a 17.2.0 apparently already cooking per their changelog but it isn't on cocoapods yet so I couldn't fetch it

I allow the facebook-ios-sdk version to slip for patch versions now but they have a warning about Swift/Obj-C compat on their repo indicating that minor versions may break compiles, so I couldn't let the semver range slip minors. They have broken us before 🤷 

Fixes #555


mikehardy:
Fixed the various deprecation and lint things I saw while examining the CI output
Note that one was not fixable as it appears to be an upstream issue, logged here https://github.com/facebook/facebook-ios-sdk/issues/2476

PR #565, opened 2024-10-12
fix(example app): use boolean values for `FacebookAutoLogAppEventsEnabled` and `FacebookAdvertiserIDCollectionEnabled`

### LINKED ISSUE #564: Info.plist entries in the example app for FacebookAutoLogAppEventsEnabled and FacebookAdvertiserIDCollectionEnabled as Strings instead of Booleans
## 🐛 Bug Report
I was reviewing the Info.plist file in the example app and noticed that the following keys are set with string values "FALSE" instead of boolean values:
```
<key>FacebookAutoLogAppEventsEnabled</key>
<string>FALSE</string>
<key>FacebookAdvertiserIDCollectionEnabled</key>
<string>FALSE</string>
```

According to Meta's documentation, these keys should be of boolean type:
[https://developers.facebook.com/docs/app-events/getting-started-app-events-ios#disable-advertiser-id](https://developers.facebook.com/docs/app-events/getting-started-app-events-ios#disable-advertiser-id)

I'm wondering if this was intentional or if it might be a mistake. Should these values be changed to booleans?

Thank you for your time and assistance!

yaroslavnikiforov:
Fixes #564

### DIFF
--- a/RNFBSDKExample/ios/RNFBSDKExample/Info.plist
+++ b/RNFBSDKExample/ios/RNFBSDKExample/Info.plist
@@ -76,9 +76,9 @@
     <string>fbshareextension</string>
   </array>
   <key>FacebookAutoLogAppEventsEnabled</key>
-  <string>FALSE</string>
+  <false/>
   <key>FacebookAdvertiserIDCollectionEnabled</key>
-  <string>FALSE</string>
+  <false/>
   <key>CFBundleURLTypes</key>
   <array>
     <dict>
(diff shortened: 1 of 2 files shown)


mikehardy (changes requested):
This is a bit obscure, but this is the wrong place to change this 😆 

https://github.com/thebergamo/react-native-fbsdk-next/blob/06217602c578fad1a8d26ffa59666b3a070b851f/refresh-example.sh#L114-L117

Is where to change it, then you re-run refresh-example.sh

What!? Why? Keeping the example up to date is actually one of the most irritating burdens of being a react-native module maintainer, so I automate it as much as possible - and this is how, in this repo

I think it looks pretty easy to edit there though - that is the required changes are reasonably obvious I think? Can you make them in the script instead? Many thanks

yaroslavnikiforov:
I updated my commit. But `refresh-example.sh` upgrades the react-native version and creates a bigger diff, should those changes be part of this PR?

mikehardy (approved):
I think it's fine to leave the concrete example update to another commit - I usually do two commits, one to update `refresh-example.sh` and one that then has the result of re-running it

But I believe I ran this pretty recently so it should be fine, and if I did not run it + commit the result recently then there was a reason and it means I have something else to fix - which would be out of scope for this PR

TL;DR: this is great, your intended + desired change is now in the example and in the refresh script so it's permanent and that's really the point

Thanks again!

PR #576, opened 2024-11-06
fix(yarn): upgrade yarn to 4.5.1

LukeGu-web:
Upgrade yarn to the latest version.

The interactive-tools plugin is included by default.


mikehardy:
🤔  perhaps the example has a lockfile as well? and needs to be committed for yarn to be okay with the public pull request thing ?

LukeGu-web:
> 🤔 perhaps the example has a lockfile as well? and needs to be committed for yarn to be okay with the public pull request thing ?

Thanks for pointing out the lockfile issue. I'm not sure what the public pull request thing is needed.

mikehardy (approved):
LGTM thanks

PR #587, opened 2024-12-13
docs: Deeplink ordering note added to README.md

Asad520:
**ISSUE**
My app has multiple SSO options set up. When I added Facebook SSO using this package, it worked fine for Android and some iOS devices as well. But for some iOS devices, the app navigates to the Facebook app to authenticate and switches back to my app with a deep link like `fbAppId://somePermissions%data%etc`.

After debugging for an entire day, I realized that I had put `RCTLinkinganager` condition before `FBSDKApplicationDelegate` and just changing the order fixed it.

I believe this short note would be helpful for anyone having multiple other SSOs while integrating this package

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ The `AppDelegate.m` file can only have one method for `openUrl`. If you're also 
   return NO;
 }
 ```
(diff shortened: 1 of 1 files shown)


Asad520:
@mikehardy Can you take a look at this, please?

mikehardy (approved):
seems legit :-) - thank you

PR #589, opened 2024-12-17
docs(example): alter refresh_example to use fb prefix in fb_login_protocol_scheme

### LINKED ISSUE #586: Difference between example and Meta guide of fb_login_protocol_scheme
In the Meta guide, it says that the fb_login_protocol_scheme should have the `fb` prefix.

```xml
<string name="facebook_app_id">1234</string>
<string name="fb_login_protocol_scheme">fb1234</string>
<string name="facebook_client_token">56789</string>
```

https://developers.facebook.com/docs/facebook-login/android/

But in the example the fb_login_protocol_scheme does not have the prefix

https://github.com/thebergamo/react-native-fbsdk-next/blob/v13.3.0/RNFBSDKExample/android/app/src/main/res/values/strings.xml#L6

which one is the correct?

mikehardy:
- Fixes #586 

Note that this is not complete, there have been some other direct changes to the example that must be incorporated into the refresh_example.sh script so they aren't lost

- https://github.com/thebergamo/react-native-fbsdk-next/commit/1cfb158e0c2bff16dc02d06d771692b11fae6299

After altering refresh-example to make that change, then refresh-example should be committed one more time, and then the results of a run should be committed (with all the updated example files...)


mikehardy:
in fact https://github.com/thebergamo/react-native-fbsdk-next/commit/1cfb158e0c2bff16dc02d06d771692b11fae6299 did a good job - it updated the refresh-example script at the same time

So I'm running it again now just to update things and verify this change but then it should be good to go

mikehardy:
Something appears to be broken in the example when it moves from react-native 0.75.x to 0.76.x and I haven't quite figured out what.

Relates to native module auto-loading as a likely root cause, and symptom is that the example app fails on boot with the Settings native module being null / inaccessible. Reproduces on android with true or false newArchEnabled in gradle.properties.

PR #595, opened 2025-01-15
feat(android): adopt facebook-android-sdk v18.0.0

hamadmarhoon:
This pull request updates the Android Facebook SDK to version 18. 
- Reason: SDK v18 introduces better support for Google Play Billing Library v5 - v7
- Tested: Verified on both Android and iOS using the example app.
- No breaking changes observed.

### DIFF
--- a/RNFBSDKExample/ios/Podfile.lock
+++ b/RNFBSDKExample/ios/Podfile.lock
@@ -1,21 +1,21 @@
 PODS:
   - boost (1.84.0)
   - DoubleConversion (1.1.6)
-  - FBAEMKit (17.4.0):
-    - FBSDKCoreKit_Basics (= 17.4.0)
+  - FBAEMKit (18.0.0):
+    - FBSDKCoreKit_Basics (= 18.0.0)
   - FBLazyVector (0.76.5)
-  - FBSDKCoreKit (17.4.0):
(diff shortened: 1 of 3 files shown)


mikehardy:
Hey there! Awesome, this looks right on for facebook-android-sdk
I see ios had a v18 release as well - https://github.com/facebook/facebook-ios-sdk/releases
You mention you checked that iOS works but I don't see the podspec dep for the facebook-ios-sdk in this PR, did you test it but not commit it? 

https://github.com/thebergamo/react-native-fbsdk-next/blob/4df764f6b883ff3092ea451033cf2337efa42743/react-native-fbsdk-next.podspec#L4

hamadmarhoon:
Hello @mikehardy! I wasn’t aware that the iOS SDK had an update as well and initially intended this PR to update only the Android SDK. However, I’ve now gone ahead and updated the iOS SDK version as well. Please let me know if there’s anything else I should adjust. Thanks again for reviewing this!

mikehardy (approved):
looks like what I'd expect, thank you!

imki123:
I couldn’t build on iOS, but after deleting all the fbsdk-related code from the `Podfile.lock` and running pod install, everything worked fine!

PR #645, opened 2025-10-09
fix(ios): native crashes when using native props

habovh:
A critical issue was introduced by #505 that breaks usage of any of the following props on iOS platform:
- `permissions`
- `defaultAudience`
- `nonceIOS`
- `loginTrackingIOS`
- `tooltipBehaviorIOS`

Attempting to use any of these props will trigger the following **runtime** native error on mount:
```log
Thread 1: "-[RCTFBSDKLoginButtonView setPermissions:]: unrecognized selector sent to instance 0x133e15610"
```

This happens because the view exposed by native is not an instance of `FBSDKLoginButton` anymore, but a new wrapper `RCTFBSDKLoginButtonView` introduced in #505. Therefore, in the various `RCT_EXPORT_VIEW_PROPERTY` and `RCT_CUSTOM_VIEW_PROPERTY` macros, the `view` references point to the wrapper `RCTFBSDKLoginButtonView` which does not implement the properties itself.

This commit exposes the underlying `FBSDKLoginButton` property on the custom `RCTFBSDKLoginButtonView` wrapper, and re-exports these props as custom properties with the correct viewClass for correct static type checking.

> [!WARNING]
> I am not proficient in Objective-C. This should be extensively tested and checked as I might have made mistakes. Do *not* assume every use case has been tested!

### DIFF
--- a/ios/RCTFBSDK/login/RCTFBSDKLoginButtonManager.m
+++ b/ios/RCTFBSDK/login/RCTFBSDKLoginButtonManager.m
@@ -41,23 +41,32 @@ RCT_EXPORT_MODULE(RCTFBLoginButton)
 
 RCT_EXPORT_VIEW_PROPERTY(onChange, RCTBubblingEventBlock)
 
-RCT_EXPORT_VIEW_PROPERTY(permissions, NSStringArray)
-
-RCT_EXPORT_VIEW_PROPERTY(defaultAudience, FBSDKDefaultAudience)
-
-RCT_CUSTOM_VIEW_PROPERTY(nonceIOS, NSString, FBSDKLoginButton)
+RCT_CUSTOM_VIEW_PROPERTY(permissions, NSStringArray, RCTFBSDKLoginButtonView)
 {
-    [view setNonce:json ? json : nil];
+    [view.loginButton setPermissions:json ? json : nil];
 }
 
-RCT_CUSTOM_VIEW_PROPERTY(loginTrackingIOS, NSString, FBSDKLoginButton)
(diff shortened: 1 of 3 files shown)


habovh:
This is not stale, but nobody is taking care of it. This PR very much needs to be merged.

maherzaidoune (approved):

habovh:
Still relevant.

dotnich-io:
Still relevant!

mikehardy:
May be relevant but has no PR and "this commit" referenced in the issue doesn't exist.

PRs welcome

habovh:
@mikehardy what do you mean? This is a PR, with suggested changes that effectively address the issue at hand —using in production on my end without issues.
The referenced *PR* (not commit) exists and has been merged.

mikehardy:
Ah 🤦 - sorry ripping through notifications too quickly, thought it was an issue not PR, lemme look

mikehardy (approved):
Code looks good, approach looks reasonable - just one note:

> > [!WARNING]
> > I am not proficient in Objective-C. This should be extensively tested and checked as I might have made mistakes. Do *not* assume every use case has been tested!

I do not have the free time to devote to this so I have done no testing. The limit of the volunteer time I'll donate to this effort is shepherding it to merge and release, if there are problems, I'll need the help of some interested party to test more + post a PR to repair. Thanks for your understanding

mikehardy:
@habovh getting it published was the hardest part after npmjs tightened their security to make software supply chain attacks more difficult with "Trusted Publish" - bit of a battle to make that work but it's released now, thanks for the PR and sorry for the confusion earlier, cheers

habovh:
@mikehardy hey no worries mate, thanks for the follow up and for approving the PR!

